PR #11, opened 2022-10-19
test: add big file test coverage

alanshaw:
This PR adds a helper `Builder` class which adds data to R2 buckets as you'd expect it to be done by dotStorage APIs.

It is used in tests and I also added a CLI tool allowing you to populate buckets for testing locally.


vasco-santos (approved):
LGTM

PR #15, opened 2022-10-20
fix: temporarily limit the size of the response until memory leak is resolved

alanshaw:
This should allow us to get freeway back into production while we fix the memory leak.

### DIFF
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -169,13 +169,10 @@ export function withMaxContentLength (maxLength, handler) {
    */
   return async (request, env, ctx) => {
     const response = await handler(request, env, ctx)
-    if (!response.headers.has('Content-Length')) return response
-
-    const contentLength = parseInt(response.headers.get('Content-Length') || '0')
+    const contentLength = parseInt(response.headers.get('Content-Length') ?? '0')
     if (contentLength > maxLength) {
       throw new HttpError('Content too big', { status: 403 })
     }
-
     return response
   }
 }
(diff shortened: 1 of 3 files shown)


src/middleware.js
```diff
@@ -155,3 +155,27 @@ export function withVersionHeader (handler) {
...
+ */
+export function withMaxContentLength (maxLength, handler) {
+  /**
+   * @type {import('@web3-storage/gateway-lib').Handler<import('@web3-storage/gateway-lib').Context>}
+   */
+  return async (request, env, ctx) => {
+    const response = await handler(request, env, ctx)
+    if (!response.headers.has('Content-Length')) return response
+
+    const contentLength = parseInt(response.headers.get('Content-Length') || '0')
```
olizilla, line 174: With the default value being provided, we can just collapse this to
```suggestion
    const contentLength = parseInt(response.headers.get('Content-Length') ?? '0')
```

olizilla (approved):
LGTM. When the tests say it works it's good to go.

vasco-santos (approved):
LGTM

src/middleware.js
alanshaw: Hmm, I thought it returned `null` and `??` only work with `undefined`?

alanshaw: oh it's nullish...

alanshaw: 👍 thanks!

PR #16, opened 2022-10-20
fix: remove max Content-Length middleware

alanshaw:
Also removes the memory budget code (which has no bearing on actual memory usage!).


src/lib/blockstore.js
```diff
@@ -13,7 +13,8 @@ import { BlockBatch } from './block-batch.js'
  * @typedef {import('../bindings').R2Bucket} R2Bucket
  */
 
-const MAX_BLOCK_LENGTH = 1024 * 1024 * 4
+// 2MB (max safe libp2p block size) + typical block header length + some leeway
+const MAX_ENCODED_BLOCK_LENGTH = (1024 * 1024 * 2) + 39 + 61
```
alanshaw, line 17: We don't know how much to read from the CAR, but in theory no single block should be bigger than this.

src/lib/block-batch.js
```diff
@@ -1,35 +1,42 @@
...
   }
 
   next () {
-    const offsets = this.#offsets // .sort((a, b) => a - b)
-    if (!offsets.length) return
+    const queue = this.#queue
+    if (!queue.length) return []
     const batch = []
-    let last = offsets[0]
+    let last = queue[0]
```
olizilla, line 28: last is first item in queue? i guess it's the last from the previous batch, but it's a little confusing that we take the head of the queue and call it last.

olizilla: also why not shift this guy off the queue and add him to the batch immediately, otherwise our first comparison is always with itself!

src/lib/block-batch.js
```diff
@@ -1,35 +1,42 @@
...
     while (true) {
-      const offset = offsets.shift()
-      if (!offset) break
-      if (offset - last >= MAX_BYTES_BETWEEN) {
-        offsets.unshift(offset) // not in this batch, return to pile
+      const item = queue.shift()
+      if (!item) break
+      if (item.carCid.toString() !== last.carCid.toString() || item.offset - last.offset >= MAX_BYTES_BETWEEN) {
+        queue.unshift(item) // not in this batch, return to pile
```
olizilla, line 33: minor but could skip a gozzy shift/unshift operation and invert the logic to make it more positive... like test for the condition we want to be true

```suggestion
```suggestion
      const item = queue.at(0)
      if (!item) break
      if (item.carCid.toString() === last.carCid.toString() && item.offset - last.offset < MAX_BYTES_BETWEEN) {
        batch.push(queue.shift(item)) // winner! remove from queue and add to item
``````

File: src/lib/blockstore.js [hunk 4]
```diff
@@ -109,69 +108,62 @@ export class BatchingR2Blockstore extends R2Blockstore {
...
+      const batch = batcher.next()
+      if (!batch.length) break
+
+      batch.sort((a, b) => a.offset - b.offset)
+
+      const { carCid } = batch[0]
+      const carPath = `${carCid}/${carCid}.car`
+      const range = {
+        offset: batch[0].offset,
+        length: batch[batch.length - 1].offset - batch[0].offset + MAX_ENCODED_BLOCK_LENGTH
```
olizilla, line 126:
```suggestion
        length: batch.at(-1).offset - batch[0].offset + MAX_ENCODED_BLOCK_LENGTH
```

src/lib/blockstore.js
```diff
@@ -109,69 +108,62 @@ export class BatchingR2Blockstore extends R2Blockstore {
...
-            blocks.forEach(b => b.resolve())
-          }
-          return
+      console.log(`fetching ${batch.length} blocks from ${carCid} (${range.length} bytes @ ${range.offset})`)
+      const res = await this._dataBucket.get(carPath, { range })
+      if (!res) {
+        for (const blocks of pendingBlocks.values()) {
+          blocks.forEach(b => b.resolve())
         }
+        return
```
olizilla, line 135: plz comment to explain this! From convo this is to furfil the contract of returning undefined if we dont have the block

src/lib/blockstore.js
```diff
@@ -109,69 +108,62 @@ export class BatchingR2Blockstore extends R2Blockstore {
...
+            blocks.forEach(b => b.resolve(block))
+            pendingBlocks.delete(key)
+          }
+        } catch {
+          break
+        }
       }
-      console.log(`finished processing batch for ${carCid}`)
-      console.log(`${batchBlocks.size} blocks remain`)
+      reader.cancel()
```
olizilla, line 165: y dis?

olizilla (approved):
Is good! Minor notes inline!

PR #24, opened 2023-02-28
feat: set max number of car cids to resolve

vasco-santos:
Sets max number of car CIDs to resolve based on issue and numbers suggested in https://github.com/web3-storage/freeway/issues/22

Later on, once bigger chunks changes is largely propagated, we can consider increasing this. Also, going to create an issue to add logging to `freeway` so that we can log number of car CIDs when we throw this error back so that we can have better visibility on these numbers

### DIFF
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -72,6 +72,13 @@ export function withCarCids (handler) {
       throw new HttpError('missing origin CAR CID(s)', { status: 400 })
     }
 
+    // Cloudflare currently sets a limit of 1000 sub-requests within the worker context
+    // If we have a given root CID splitted across hundreds of CARs, freeway will hit
+    // the sub-requests limit and not serve content anyway
+    if (carCids.length > Number(env.MAX_CAR_CIDS_TO_RESOLVE)) {
+      throw new HttpError('number CAR CIDs is too large to resolve', { status: 501 })
+    }
+
     return handler(request, env, { ...ctx, carCids })
   }
 }
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -40,6 +40,11 @@ export function withCarCids (handler) {
     if (!ctx.dataCid) throw new Error('missing data CID')
     if (!ctx.searchParams) throw new Error('missing URL search params')
 
+    // Cloudflare currently sets a limit of 1000 sub-requests within the worker context
+    // If we have a given root CID splitted across hundreds of CARs, freeway will hit
+    // the sub-requests limit and not serve content anyway
+    const maxShards = env.MAX_SHARDS ? parseInt(env.MAX_SHARDS) : 250
+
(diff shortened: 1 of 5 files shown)


src/middleware.js
```diff
@@ -72,6 +72,13 @@ export function withCarCids (handler) {
       throw new HttpError('missing origin CAR CID(s)', { status: 400 })
     }
 
+    // Cloudflare currently sets a limit of 1000 sub-requests within the worker context
+    // If we have a given root CID splitted across hundreds of CARs, freeway will hit
+    // the sub-requests limit and not serve content anyway
+    if (carCids.length > Number(env.MAX_CAR_CIDS_TO_RESOLVE)) {
+      throw new HttpError('number CAR CIDs is too large to resolve', { status: 501 })
+    }
```
alanshaw, line 80: Move into the while loop above for quicker response?

wrangler.toml
```diff
@@ -27,6 +27,9 @@ r2_buckets = [
 [env.production.build]
 command = "npm run build"
 
+[env.production.vars]
+MAX_CAR_CIDS_TO_RESOLVE = "250"
```
alanshaw, line 31: It's tricky because a DAG might be split across 999 CARs but because of the path you might be able to extract `/smol.json` with the last sub-request.

alanshaw: With our 10MB CAR shards it'll restrict freeway from serving anything that's over 2.5GB

src/middleware.js
```diff
@@ -72,6 +72,13 @@ export function withCarCids (handler) {
       throw new HttpError('missing origin CAR CID(s)', { status: 400 })
     }
 
+    // Cloudflare currently sets a limit of 1000 sub-requests within the worker context
+    // If we have a given root CID splitted across hundreds of CARs, freeway will hit
+    // the sub-requests limit and not serve content anyway
+    if (carCids.length > Number(env.MAX_CAR_CIDS_TO_RESOLVE)) {
+      throw new HttpError('number CAR CIDs is too large to resolve', { status: 501 })
```
alanshaw, line 79:
```suggestion
    const maxShards = env.MAX_SHARDS ? parseInt(env.MAX_SHARDS) : 250
    if (carCids.length > maxShards) {
      throw new HttpError('request exceeds maximum DAG shards', { status: 501 })
```

alanshaw (changes requested):

wrangler.toml
vasco-santos: I understand the disadvantage for these use cases. However, at this point looks like the priority should be to reduce R2 ops and avoid potential error states mentioned in issue (break in mid stream without feedback for user, break after worker run out of sub-requests in general). When we have bigger sub-requests limit and better R2 perf we can increase these.

We can tweak more the number if you feel like there is a better compromise. Thoughts?

alanshaw (approved):

PR #27, opened 2023-03-15
fix: allow hash character in directory listings

alanshaw:
Pulls in the new `gateway-lib` with the fix for allowing hash character in directory listings.

Also updates all dependencies, which includes a change in dagula to switch from our unixfs exporter fork back to `ipfs-unixfs-exporter` (since the parallel block fetching from the fork was backported).

`ipfs-unixfs-exporter` now uses `.subarray()` everywhere (no buffer copies) which is rad, but it also caused our tests to fail which was tricky to track down the cause. I've commented heavily in the place where the fix is.

### DIFF
--- a/scripts/r2-put.js
+++ b/scripts/r2-put.js
@@ -11,7 +11,7 @@ import { Builder } from '../test/helpers.js'
 
 const bucketNames = ['CARPARK', 'SATNAV', 'DUDEWHERE']
 const buckets = bucketNames.map(b => new R2Bucket(new FileStorage(`./.mf/r2/${b}`)))
-const builder = new Builder(...buckets)
+const builder = new Builder(buckets[0], buckets[1], buckets[2])
 
 const paths = process.argv.slice(2).filter(p => p !== '--no-wrap')
 const files = await getFilesFromPath(paths)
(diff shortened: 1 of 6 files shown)


scripts/r2-put.js
```diff
@@ -11,7 +11,7 @@ import { Builder } from '../test/helpers.js'
 
 const bucketNames = ['CARPARK', 'SATNAV', 'DUDEWHERE']
 const buckets = bucketNames.map(b => new R2Bucket(new FileStorage(`./.mf/r2/${b}`)))
-const builder = new Builder(...buckets)
+const builder = new Builder(buckets[0], buckets[1], buckets[2])
```
alanshaw, line 14: Changed to stop typescript complaining...

src/lib/blockstore.js
```diff
@@ -156,7 +156,37 @@ export class BatchingR2Blockstore extends R2Blockstore {
           const blocks = pendingBlocks.get(key)
           if (blocks) {
             // console.log(`got wanted block for ${blockHeader.cid}`)
-            const block = { cid: blockHeader.cid, bytes }
+            const block = {
+              cid: blockHeader.cid,
+              // FML, this took _so_ long to figure out:
```
vasco-santos, line 161: oh man! glad you ended up finding it out

vasco-santos (approved):

PR #45, opened 2023-06-05
feat: upgrade cardex

alanshaw:
Upgrade to the new version of cardex, which has a web stream-y API similar to https://github.com/ipld/js-unixfs.


vasco-santos (approved):
lgtm

PR #46, opened 2023-06-05
feat: use rollup indexes

alanshaw:
This PR allows Freeway to use rollup indexes. Rollups are multi-index indexes (see https://github.com/alanshaw/cardex#multi-index-index) that specify which blocks are in which CAR file(s) for a given DAG.

Freeway checks DUDEWHERE for a rollup index at `bafyROOT/.rollup.idx`. If it finds one, it'll use it (and no other indexes) to obtain index informations for all the blocks.

### DIFF
--- a/test/index.spec.js
+++ b/test/index.spec.js
@@ -102,8 +102,6 @@ describe('freeway', () => {
     const input = [{ path: 'sargo.tar.xz', content: randomBytes(609261780) }]
     const { dataCid, carCids } = await builder.add(input)
 
-    await builder.rollup(dataCid, carCids)
-
     // remove the the CAR CIDs from DUDEWHERE so that only the rollup index can
     // be used to satisfy the request.
     const bucket = await miniflare.getR2Bucket('DUDEWHERE')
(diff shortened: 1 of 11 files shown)


test/index.spec.js
```diff
@@ -97,4 +97,17 @@ describe('freeway', () => {
     const output = new Uint8Array(await res.arrayBuffer())
     assert.doesNotReject(CarReader.fromBytes(output))
   })
+
+  it('should use a rollup index', async () => {
+    const input = [{ path: 'sargo.tar.xz', content: randomBytes(609261780) }]
+    const { dataCid, carCids } = await builder.add(input)
+
+    await builder.rollup(dataCid, carCids)
```
vasco-santos, line 105: looking into test `should get a big file` without the rollup, it is exactly the same, which means we are not validating it comes from rollup? 

could we wire a call count or something like that that we could try to leverage here?

vasco-santos (approved):
Code looks great to me! It would be nice to have better test vector for it

test/index.spec.js
alanshaw: I was thinking to remove the non-rollup indexes, so that the only way it can successfully be served would be via rollup.

PR #62, opened 2023-07-10
feat: cache index data

alanshaw:
Caches satnav data in cloudflare's cache.

### DIFF
--- a/test/index.spec.js
+++ b/test/index.spec.js
@@ -28,7 +28,8 @@ describe('freeway', () => {
       modules: true,
       r2Buckets: bucketNames,
       // r2Persist: true
-      kvNamespaces: ['BLOCKLY']
+      kvNamespaces: ['BLOCKLY'],
+      usageModel: 'unbound'
     })
 
     const buckets = await Promise.all(bucketNames.map(b => miniflare.getR2Bucket(b)))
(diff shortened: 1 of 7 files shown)


test/index.spec.js
```diff
@@ -28,7 +28,8 @@ describe('freeway', () => {
       modules: true,
       r2Buckets: bucketNames,
       // r2Persist: true
-      kvNamespaces: ['BLOCKLY']
+      kvNamespaces: ['BLOCKLY'],
+      usageModel: 'unbound'
```
olizilla, line 32: just to make test more like prod? or was it necessary for some wild reason?

olizilla (approved):
is it possible to test the cache behaviour? Like test a fetch, then delete from the test bucket, then fetch again and assert is retrieved? or some other way to prove that the cache is doing something?

test/index.spec.js
alanshaw: In one of the tests we read 65 indexes, and with bound usage it took us over the limit of 50.

vasco-santos (approved):

test/index.spec.js
alanshaw: It also makes it more like prod ;)

PR #68, opened 2023-07-24
feat: serve CAR blocks

alanshaw:
This PR adds a new handler that allows freeway to serve CAR files directly from R2 (CARPARK).

Requesting a CID from the gateway with the CAR codec `0x0202` will invoke this new handler.

Obviously CARs are not served in the same way as regular IPFS data, since we can serve the data directly from an R2 bucket. To deal with this difference, there's a new middlewares that detects the CAR codec and calls the correct handler, passing through to the regular handlers if the CAR codec is not detected.

The handler supports `HEAD` requests, allowing you to get the file size of the item you're requesting. It also supports the HTTP `Range` header, allowing you to extract byte ranges from within a CAR, which you might do if you fetch the index for the CAR.

The idea is that we can use these handlers to build a system that runs on content-claims. The flow might be something like:

1. Get [partition claim](https://github.com/web3-storage/content-claims#partition-claim) for a given content CID
2. Get [inclusion claims](https://github.com/web3-storage/content-claims#inclusion-claim) for each CAR `part` in the partition claim
3. Fetch each CARv2 index from the gateway for each `includes` CID specified in the inclusion claims (You must first fetch partition claims for each index CID to find out which CAR file they can be found in).
4. Use HTTP `Range` requests to export the whole DAG or a sub-DAG directly from the `parts`, using the index data as a guide for which blocks to extract.

Note: if exporting a full DAG you might want to just stop after (1) and import all data in all CARs into IPFS.


src/handlers/car-block.js
```diff
@@ -0,0 +1,82 @@
...
+  let range
+  if (request.headers.has('range')) {
+    try {
+      range = Http.parseRange(request.headers.get('range') ?? '')
+    } catch (err) {
+      throw new HttpError('invalid range', { status: 400, cause: err })
+    }
+  }
+
+  const obj = await env.CARPARK.get(`${dataCid}/${dataCid}.car`, { range })
```
vasco-santos, line 53: Could be cool to try this with roundabout. I think it should work and let us redirect to presigned url. 

Probably good to keep as is and track issue to attempt cost optimisation :)

test/helpers/builder.js
```diff
@@ -57,8 +58,34 @@ export class Builder {
       writer.add(entry.cid, entry.offset)
     }
     writer.close()
-    // @ts-expect-error node web stream is not web stream
-    await this.#satnav.put(`${cid}/${cid}.car.idx`, readable)
+
+    // @ts-expect-error
+    const indexBytes = concat(await collect(readable))
+    await this.#satnav.put(`${cid}/${cid}.car.idx`, indexBytes)
```
vasco-santos, line 64: nit: do we really need to buffer it in memory now?

vasco-santos: if so, could be cool to state reason for this in comment

vasco-santos (approved):
LGTM! ✨

test/helpers/builder.js
alanshaw: Oh, no we don't need to here...I was just copy pasta 🤪

src/handlers/car-block.js
alanshaw: Hmm, yes but we probably don't want to give out those URLs to anyone via the gateway. Someone will try to use/abuse and they also expire...

vasco-santos: So the usage would be the same as here? it would be a redirect that response will be read right away right? So, expiration would not be a problem, and if we perform request to `roundabout` within the worker and redirect to presigned URL then user will not even know how URL was created.

Use/abuse would be the same as via freeway no. So, rate limits is really the solution we seem to have to avoid abuse

PR #75, opened 2023-08-14
feat: add generate rollup script

alanshaw:
Adds a script that generate a rollup index for a given root CID.

### DIFF
--- a/scripts/rollup.js
+++ b/scripts/rollup.js
@@ -1,0 +1,178 @@
+/**
+ * Generate a rollup index, given a root CID.
+ *
+ * Reads DUDEWHERE to find all CAR files the CID's DAG may be found in, and
+ * reads each SATNAV index to generate a rollup index.
+ *
+ * Finally the rollup index is written to DUDEWHERE bucket.
+ *
+ * All is performed over S3 HTTP API.
+ *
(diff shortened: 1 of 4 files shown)


scripts/rollup.js
```diff
@@ -0,0 +1,178 @@
...
+  const chunks = []
+  // @ts-expect-error
+  for await (const chunk of readable) {
+    chunks.push(chunk)
+  }
+  const index = Buffer.concat(chunks)
+
+  console.log(`⏳ Writing index (${index.length} bytes)...`)
+  const key = `${root}/.rollup.idx`
+  const cmd = new PutObjectCommand({ Bucket: dudewhereBucket, Key: key, Body: index })
```
olizilla, line 102: minor: Could convert your `ReadableStream` to a node `Readable` and pass that as `Body`

https://github.com/awslabs/smithy-typescript/blob/f785a574d75899c5fa91f8c5b4ac5bbae2564ed6/packages/types/src/streaming-payload/streaming-blob-payload-input-types.ts#L56C90-L56C98

olizilla (approved):
👍

olizilla:
I think we have a lot of useful scripts in a few repos... i suspect the team doesn't know about them. Maybe add a note to the README? or a notion? or a how-to repo?

PR #93, opened 2023-12-06
feat: content claims reads by default with fallback for old index sources

vasco-santos:
Makes content claims used by default. If no index was found from content claims, the handler is returned wrapped with the old `withIndexerSources` and `withDagula` (old)

Alternatively, we could add a new middleware and keep there the old ones. We could inspect in these middlewares if we have a dagula instance already, but I don't think that is a good way to go as we in the future can just drop this fallback

### DIFF
--- a/test/index.spec.js
+++ b/test/index.spec.js
@@ -190,6 +190,30 @@ describe('freeway', () => {
     assert(equals(input[0].content, output))
   })
 
+  it('should use content claims by default', async () => {
+    const input = [{ path: 'sargo.tar.xz', content: randomBytes(MAX_CAR_BYTES_IN_MEMORY + 1) }]
+    // no dudewhere or satnav so only content claims can satisfy the request
+    const { dataCid, carCids, indexes } = await builder.add(input, {
+      dudewhere: true,
+      satnav: true
+    })
+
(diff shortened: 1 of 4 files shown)


test/index.spec.js
```diff
@@ -190,6 +190,30 @@ describe('freeway', () => {
...
+    // @ts-expect-error nodejs ReadableStream does not implement ReadableStream interface correctly
+    const claims = await generateClaims(claimsService.signer, dataCid, carCids[0], res.body, indexes[0].cid, indexes[0].carCid)
+    claimsService.setClaims(claims)
+
+    const res1 = await miniflare.dispatchFetch(`http://localhost:8787/ipfs/${dataCid}/${input[0].path}`)
+    if (!res1.ok) assert.fail(`unexpected response: ${await res1.text()}`)
+
+    const output = new Uint8Array(await res1.arrayBuffer())
+    assert(equals(input[0].content, output))
+    assert.notEqual(claimsService.getCallCount(), 0)
```
alanshaw, line 214: You could have a `claimsService.resetCallCount()` that is called `beforeEach` and then you could assert on the real value.

alanshaw (approved):

vasco-santos:
Waiting on 🟢  light from Oli to get this merged and deployed

PR #100, opened 2024-05-02
feat: add blob handler

alanshaw:
Adds `GET/blob/:multihash`

refs https://github.com/w3s-project/w3infra/pull/363

### DIFF
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -11,6 +11,7 @@ import { MultiCarIndex, StreamingCarIndex } from './lib/dag-index/car.js'
 import { CachingBucket, asSimpleBucket } from './lib/bucket.js'
 import { MAX_CAR_BYTES_IN_MEMORY, CAR_CODE } from './constants.js'
 import { handleCarBlock } from './handlers/car-block.js'
+import { handleBlob } from './handlers/blob.js'
 
 /**
  * @typedef {import('./bindings.js').Environment} Environment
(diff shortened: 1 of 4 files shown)


src/middleware.js
```diff
@@ -52,6 +53,22 @@ export function withCarHandler (handler) {
...
+/**
+ * Middleware that will serve blobs if `/blob/:multihash` is found in the path.
+ * If path does not start with `/blob/` it delegates to the next middleware.
+ *
+ * @type {import('@web3-storage/gateway-lib').Middleware<import('@web3-storage/gateway-lib').Context, import('@web3-storage/gateway-lib').Context, Environment>}
+ */
+export function withBlobHandler (handler) {
+  return async (request, env, ctx) => {
+    const url = new URL(request.url)
+    if (!url.pathname.startsWith('/blob/')) {
```
Gozala, line 65: Do we even need the `/blob/` interface as in is there reason why we would not use RAW cid handler instead ? If we did query content claims for raw blocks we would effectively get pointer to the file covering it's full content would not we ?

Gozala (approved):
Looks good to me, although I do wonder if we should instead overload raw block read interface instead of adding another dependency on carpark

src/middleware.js
alanshaw: Agree, lets move this into the raw handler.

vasco-santos: yes, I also agree, and think actually having `/blob` will just make everything more complicated on the reads side

PR #101, opened 2024-05-13
feat: support byte range for raw block requests

alanshaw:
Pulls in latest gateway-lib and dagula to allow byte range requests to raw blocks (`?format=raw`).

supersedes https://github.com/web3-storage/freeway/pull/100


test/helpers/builder.js
```diff
@@ -173,6 +170,15 @@ export class Builder {
   }
 }
 
+/** @param {import('multiformats').Link} cid */
+export const toCarKey = cid => `${cid}/${cid}.car`
+
+/** @param {import('multiformats').MultihashDigest} digest */
+export const toBlobKey = digest => {
+  const digestString = base58btc.encode(digest.bytes)
+  return `${digestString}/${digestString}.blob`
```
Gozala, line 179: Nit: I've been thinking whether doing something like `blob/${digestSstring}.blob` would be a better option ? At least visually it would be easier to distinguish that looking at car vs blob hashes and potentially other things in the future.

test/helpers/content-claims.js
```diff
@@ -122,41 +122,51 @@ export const generateClaims = async (signer, dataCid, carCid, carStream, indexCi
...
-              (`http://localhost/${carCid}/${carCid}.car`)
-            ],
-            range: { offset: blockOffset, length: blockLength }
-          }
-        })
-
         const blocks = claims.get(cid) ?? []
-        // @ts-expect-error
-        blocks.push(await encode(invocation))
+        const location = new URL(`https://w3s.link/ipfs/${carCid}?format=raw`)
```
Gozala, line 134: I think we should use raw CIDs here as opposed to CAR CIDs with `format=raw` should not we ?

Gozala (approved):
Looks good, I do feel however that following two changes would be a recommended, but not required:

1. Location claims should probably use raw CIDs in the URL as opposed to CAR CIDs. We already had mistagged CAR content which this would avoid. I do have to admit that I don't fully understanding if such a change would invalidate assumptions of other components, in which case it will not be worth it, yet it would be good idea to document that in the code comments.
2. I think if we have raw CID we should not require additional signals in form of query param or headers. This probably could be a followup improvement

test/helpers/builder.js
alanshaw: @vasco-santos WDYT? We originally used `cid/cid.car` so that we could put indexes next to the data, like `cid/cid.car.idx` but we never did that and used a different bucket instead. So I'm not too bothered about having the extra "directory" per upload.

This would also be a good way to segregate new blob data from the old.

src/lib/dag-index/api.ts
```diff
@@ -1,11 +1,31 @@
...
+  site: {
+    location: URL[],
+    range: Required<ByteRange>
+  }
+}
+
+export type IndexEntry = NotLocatedIndexEntry | LocatedIndexEntry
+
 export interface Index {
   get (c: UnknownLink): Promise<IndexEntry|undefined>
```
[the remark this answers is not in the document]
alanshaw: Right, I mean I wrote all this before I even knew about `Result` - whilst I don't disagree (I like the type information `Result` gives us for errors) I think we should do a transition to it separately to this PR.

src/lib/dag-index/api.ts
```diff
@@ -1,11 +1,31 @@
...
+ */
+export interface LocatedIndexEntry {
+  digest: MultihashDigest
+  site: {
+    location: URL[],
+    range: Required<ByteRange>
+  }
+}
+
+export type IndexEntry = NotLocatedIndexEntry | LocatedIndexEntry
```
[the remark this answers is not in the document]
alanshaw: 🤷 if you're ok with that I can get behind it. I was thinking lost implies that you had it at some point...

src/lib/blockstore.js
```diff
@@ -32,6 +37,21 @@ export class R2Blockstore {
...
+      // if host is "w3s.link" then content can be found in CARPARK
+      const url = entry.site.location.find(l => l.hostname === 'w3s.link')
+      // TODO: allow fetch from _any_ URL
+      if (!url) return
+
+      const link = Link.parse(url.pathname.split('/')[2])
+      const digestString = base58btc.encode(link.multihash.bytes)
+      const key = `${digestString}/${digestString}.blob`
+      const res = await this._dataBucket.get(key, { range: entry.site.range })
+      if (!res) return
```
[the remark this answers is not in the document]
alanshaw: I wrote all this before I even knew about `Result`. I don't disagree (I like the type information `Result` gives us for errors) but I think we should do a transition to it separately to this PR.

test/helpers/builder.js
vasco-santos: We also do this keying with pathing to better optimize S3 Rate limits. Re R2 Rate limits, this was not rate limited like S3, but last time we talked (6 months + ago), they would be looking into having that possibility on their end. 

So, I would push back on doing this change, and prioritize to finish what we need on the reads interface to swap the bucket

vasco-santos (approved):

test/helpers/builder.js
vasco-santos: Also, we have different tables in Dynamo for Blobs and Store allocations. Therefore, if we ever need to go through the whole list, we can rely on dynamo instead of a list with prefix from bucket

src/index.js
```diff
@@ -37,30 +39,47 @@ export default {
...
-    return await handleBlock(request, env, ctx)
+/**
+ * @type {import('@web3-storage/gateway-lib').Middleware<BlockContext & UnixfsContext & IpfsUrlContext, BlockContext & UnixfsContext & IpfsUrlContext, Environment>}
+ */
+export function withFormatRawHandler (handler) {
+  return async (request, env, ctx) => {
+    const { headers } = request
+    const { searchParams } = ctx
+    if (!searchParams) throw new Error('missing URL search params')
+    if (searchParams.get('format') === 'raw' || headers.get('Accept')?.includes('application/vnd.ipld.raw')) {
```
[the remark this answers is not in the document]
alanshaw: This middleware is designed to only handle the request if `?format=raw` is in the query string or `Accept: application/vnd.ipld.raw` is in the HTTP headers.

I will make this handler also handle requests when the CID codec is raw, but it requires some other changes to keep parity with what the unixfs handler does when a raw block is requested (mainly content type sniffing).

PR #103, opened 2024-05-17
feat: fetch blocks from location URLs

alanshaw:
Freeway now attempts to `fetch` data from claimed location URLs and does not rely on an R2 bucket.


vasco-santos (approved):
LGTM! ✨

PR #105, opened 2024-05-24
feat: use blob-fetcher lib

alanshaw:
Uses `@web3-storage/blob-fetcher`, `@web3-storage/public-bucket` and `multipart-byte-range` to simplify code and ensure that freeway uses the same code as hoverboard to fetch blocks via HTTP range requests informed by location content claims.

This removes the content claims code from the existing R2 blockstore so that it can be decomssioned more easily as there is clean separation between new world and old world.


src/middleware.js
```diff
@@ -56,27 +57,49 @@ export function withCarBlockHandler (handler) {
...
+      async get (cid) {
+        const res = await fetcher.fetch(cid.multihash)
+        return res.ok ? { cid, bytes: await res.ok.bytes() } : undefined
+      },
+      async stream (cid, options) {
+        const res = await fetcher.fetch(cid.multihash, options)
+        // @ts-expect-error `MINIFLARE` is not a property of `globalThis`
+        if (globalThis.MINIFLARE && res.ok) {
+          return res.ok.stream().pipeThrough(new TransformStream({
+            transform: (chunk, controller) => controller.enqueue(chunk.slice())
```
vasco-santos, line 93: can we document code path reasoning here?

vasco-santos (approved):
LGTM

src/middleware.js
alanshaw: ha, yes...it's this: https://github.com/web3-storage/freeway/blob/1c7b7a6faff866bb7c8fece427777e33e7e27416/src/lib/blockstore.js#L193-L220

PR #109, opened 2024-06-13
wip: spike on adding rate limits to freeway

travis:
Add a new middleware that checks a rate limiting service and returns a 429 if the CID is over a rate limit.

This sketches out an API for the rate limiting and accounting services suggested in https://github.com/storacha-network/RFC/pull/28

This is not ready to merge, but should probably be the starting point for this work once we all agree that this is the right shape.


fforbeck:
@travis we have this implemented via https://github.com/storacha/freeway/pull/115. I will close this PR. Happy to reopen it if needed.

PR #115, opened 2024-09-26
feat: rate limiter + unit tests + readme

fforbeck:
This PR add tests for the Rate Limiter middleware, the main changes include:

- [x] Updated README with steps and configs to run it locally
   - [x] Project Setup
   - [x] Running Tests
- [x] Added Mocha, Chai and Sinon packages to run Unit Tests
- [x] Rate Limiter
   - [x] Feature Flag (env variable)
   - [x] Fixes
      - [x] Rate Limiter middleware needs to be placed after the IPFS Parsed URL middleware, so we can access the CID 
      - [x] Refactored the `import {version} from './package.json' to be set via task and imported as ENV variable, so we can easily tweak that in the tests and don't run into import issues
   - [x] Unit Tests
     - 1. Calls next handler if no auth token is provided and rate limit is not exceeded
     - 2. Throws an error if no auth token is provided and rate limit is exceeded
     - 3. Calls next handler if auth token is present but no token metadata is found and rate limit is not exceeded
     - 4. Throws an error if auth token is present but no token metadata is found and rate limit is exceeded
     - 5. Calls next handler if auth token is present and token metadata is invalid but rate limit is not exceeded
     - 6. Throws an error if auth token is present and token metadata is invalid and rate limit is exceeded
     - 7. Calls next handler if auth token is present and token metadata is valid
   - [x] Update Github Action to execute Unit Tests


travis:
Could you say more about why Jest is useful here? We've typically used `node:test` for this project, and Mocha (https://mochajs.org/) or AVA (https://github.com/avajs/ava) on other projects - I'm not entirely sure why we've shied away from Jest, but I am a bit hesitant to add yet another testing framework to the mix for our team. Could definitely be convinced - I've used Jest in the past and it's generally pretty good - but need a bit more convincing ;-p

fforbeck:
> Could you say more about why Jest is useful here? We've typically used `node:test` for this project, and Mocha (https://mochajs.org/) or AVA (https://github.com/avajs/ava) on other projects - I'm not entirely sure why we've shied away from Jest, but I am a bit hesitant to add yet another testing framework to the mix for our team. Could definitely be convinced - I've used Jest in the past and it's generally pretty good - but need a bit more convincing ;-p

sure thing @travis 

I don’t have super strong opinions on this either, but Jest does give you some nice built-in features like mocking, spies, and coverage reports right out of the box. With Mocha (or node:test), you’d need to add additional libraries like Sinon for mocks and nyc for coverage, which Jest handles natively.

That said, if we can use Mocha alongside node:test, it should work fine, even though Mocha isn't as feature-rich in those areas. Totally open to whichever approach the team prefers! 😊

travis:
ok cool - I lean toward Mocha, but tbh I'd like @alanshaw to weigh in since he's written a lot more of this code (like, infinitely more, as this is really my first freeway contribution as well!) - Alan any preferences/thoughts on testing frameworks here?

package.json
```diff
@@ -13,13 +13,16 @@
   },
   "scripts": {
     "prepare": "npm run build",
-    "start": "npm run dev",
+    "set-version": "node ./set-version-env.js",
+    "start": "export $(npm run -s set-version) && npm run dev",
     "dev": "npm run build:debug && miniflare dist/worker.mjs --watch --debug -m --r2-persist --global-async-io --global-timers",
-    "build": "esbuild --bundle src/index.js --format=esm --sourcemap --minify --outfile=dist/worker.mjs && npm run build:tsc",
-    "build:debug": "esbuild --bundle src/index.js --format=esm --outfile=dist/worker.mjs",
+    "build": "export $(npm run -s set-version) && esbuild --bundle src/index.js --format=esm --sourcemap --minify --outfile=dist/worker.mjs && npm run build:tsc",
```
travis, line 19: question: should these just be added to `prepare` above? it would be nice to avoid duplicating this line, and seems like `set-version` can probably safely be run any time `prepare` is triggered - does that sound right to you?

travis (changes requested):
I've added a few comments, some of them blocking - I took https://conventionalcomments.org/ for a spin on this review, curious if you like it! 

I probably won't Approve this PR in general since I've never contributed to Freeway myself - @alanshaw could you take a look at this when you get a chance?

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+  const rateLimiter = env.RATE_LIMITER
+  if (!rateLimiter) {
+    console.warn('no rate limiter found')
+    return RATE_LIMIT_EXCEEDED.NO
+  }
+  const rateLimitResponse = await rateLimiter.limit({ key: cid.toString() })
+  if (rateLimitResponse.success) {
+    return RATE_LIMIT_EXCEEDED.NO
+  } else {
+    console.log(`limiting CID ${cid}`)
```
alanshaw, line 53:
```suggestion
    console.log(`rate limit exceeded: ${cid}`)
```

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+ * @returns TokenMetadata
+ */
+async function getTokenMetadata (env, authToken) {
+  const cachedValue = await env.AUTH_TOKEN_METADATA.get(authToken)
+  // TODO: we should implement an SWR pattern here - record an expiry in the metadata and if the expiry has passed, re-validate the cache after
+  // returning the value
+  if (cachedValue) {
+    return deserializeTokenMetadata(cachedValue)
+  } else {
+    const accounting = Accounting.create({ serviceURL: env.ACCOUNTING_SERVICE_URL })
```
alanshaw, line 71: Maybe pass the cache and accounting API instace to this function? In general we should avoid passing everything everywhere since it makes it harder to test.

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+  return JSON.stringify(m)
+}
+
+/**
+ *
+ * @param {Environment} env
+ * @param {import('@web3-storage/gateway-lib/handlers').CID} cid
+ * @returns {Promise<import('./constants.js').RATE_LIMIT_EXCEEDED>}
+ */
+async function checkRateLimitForCID (env, cid) {
```
alanshaw, line 43: This only ever uses one value from env - consider simply passing the API instance not the entire environment.

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+  const cachedValue = await env.AUTH_TOKEN_METADATA.get(authToken)
+  // TODO: we should implement an SWR pattern here - record an expiry in the metadata and if the expiry has passed, re-validate the cache after
+  // returning the value
+  if (cachedValue) {
+    return deserializeTokenMetadata(cachedValue)
+  } else {
+    const accounting = Accounting.create({ serviceURL: env.ACCOUNTING_SERVICE_URL })
+    const tokenMetadata = await accounting.getTokenMetadata(authToken)
+    if (tokenMetadata) {
+      await env.AUTH_TOKEN_METADATA.put(authToken, serializeTokenMetadata(tokenMetadata))
```
alanshaw, line 74: Please take a look at: https://developers.cloudflare.com/workers/runtime-apis/context/#waituntil

Do we need to wait for this as part of the request?

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+    }
+  }
+}
+
+/**
+ * @type {import('./bindings.js').RateLimits}
+ */
+const RateLimits = {
+  create: ({ env }) => ({
+    check: async (cid, options) => {
```
alanshaw, line 87: Please can we have some type information here?

src/bindings.d.ts
```diff
@@ -1,7 +1,47 @@
...
+  check: (cid: CID, options: GetCIDRequestOptions) => Promise<RateLimitExceeded>
+}
+
+export interface TokenMetadata {
+  locationClaim?: unknown // TODO: figure out the right type to use for this - we probably need it for the private data case to verify auth
+  invalid?: boolean
 }
+
+export interface RateLimits {
+  create: ({ env }: { env: Environment }) => RateLimitsService
```
alanshaw, line 36: I think we don't generally define types for like this as we'd never pass this around in code. Instead this would just be a file with an exported function called create.

src/bindings.d.ts
```diff
@@ -1,7 +1,47 @@
...
+  AUTH_TOKEN_METADATA: KVNamespace
+  FF_RATE_LIMITER_ENABLED: boolean
+}
+
+export type GetCIDRequestData = Pick<Request, 'url' | 'headers'>
+
+export type GetCIDRequestOptions = GetCIDRequestData
+
+export interface RateLimitsService {
+  check: (cid: CID, options: GetCIDRequestOptions) => Promise<RateLimitExceeded>
```
alanshaw, line 27: `RateLimitExceeded` - I don't see this defined anywhere...

alanshaw, line 27: `options: GetCIDRequestOptions` are they optional? If not then specify as `config` or `context`.

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+}
+
+/**
+ *
+ * @type {import('@web3-storage/gateway-lib').Middleware<IpfsUrlContext, IpfsUrlContext, Environment>}
+ */
+export function withRateLimits (handler) {
+  return async (request, env, ctx) => {
+    if (!env.FF_RATE_LIMITER_ENABLED) {
+      console.warn('rate limiting disabled')
```
alanshaw, line 150: Log just once at startup please not every request!

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+ */
+const Accounting = {
+  create: ({ serviceURL }) => ({
+    record: async (cid, options) => {
+      console.log(`using ${serviceURL} to record a GET for ${cid} with options`, options)
+    },
+
+    getTokenMetadata: async () => {
+      // TODO I think this needs to check the content claims service (?) for any claims relevant to this token
+      // TODO do we have a plan for this? need to ask Hannah if the indexing service covers this?
```
alanshaw, line 126: Yes eventually this will be indexing service, but you may have to use content claims temporarily. Either way I believe the current idea is for them to have the same API.

alanshaw (changes requested):
Love the docs updates 🙏 thank you.

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+  return authToken
+}
+
+/**
+ *
+ * @type {import('@web3-storage/gateway-lib').Middleware<IpfsUrlContext, IpfsUrlContext, Environment>}
+ */
+export function withRateLimits (handler) {
+  return async (request, env, ctx) => {
+    if (!env.FF_RATE_LIMITER_ENABLED) {
```
alanshaw, line 149:
```suggestion
    if (env.FF_RATE_LIMITER_ENABLED !== 'true') {
```

It'll be a string or undefined. Best to explicitly require the string "true".

alanshaw:
> > Could you say more about why Jest is useful here? We've typically used `node:test` for this project, and Mocha (https://mochajs.org/) or AVA (https://github.com/avajs/ava) on other projects - I'm not entirely sure why we've shied away from Jest, but I am a bit hesitant to add yet another testing framework to the mix for our team. Could definitely be convinced - I've used Jest in the past and it's generally pretty good - but need a bit more convincing ;-p
> 
> sure thing @travis
> 
> I don’t have super strong opinions on this either, but Jest does give you some nice built-in features like mocking, spies, and coverage reports right out of the box. With Mocha (or node:test), you’d need to add additional libraries like Sinon for mocks and nyc for coverage, which Jest handles natively.
> 
> That said, if we can use Mocha alongside node:test, it should work fine, even though Mocha isn't as feature-rich in those areas. Totally open to whichever approach the team prefers! 😊

Slightly my bad here, I tried `node --test` because it was new, but turns out it isn't great so I'd welcome a switch to something else. We also don't use it in any other repo.

I'm mildly against having 2 testing frameworks in the same repo, but at the same time I don't want to waste time here - there's a bunch of tests written now using it so...meh. The point is more that testing exists at all - I don't mind which framework we use. People like consistency though, and using the same framework(s) everywhere allows us to work more quickly and without context switches. We should establish a single/set of frameworks we're comfortable using as a team. I realise that may have changed now. @travis this is something needed to be decided for team norms.

Just a couple of notes: 1) I though jest was legacy/unmaintained? 2) We don't use jest _anywhere_ else.

package.json
fforbeck: Yes, that sounds right. I will update that. thanks

fforbeck: ah hold on, prepare task is run when you install the packages, but we can improve this code for sure.

set-version-env.js
```diff
@@ -0,0 +1,13 @@
...
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = dirname(__filename)
+
+const packageJsonPath = join(__dirname, 'package.json')
+const packageJson = JSON.parse(readFileSync(packageJsonPath, 'utf-8'))
+
+const { version } = packageJson
+
+console.log(`VERSION=${version}`)
```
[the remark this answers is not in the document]
fforbeck: It is missing the `process.env.VERSION=version` - I will add it. The idea is to use this script to read the version from package json and set it into the `VERSION` env var instead of importing the `package.json` file directly in the middleware.js - which was causing several failures in the tests because it couldn't resolve the `.json` import. The version is used by the middleware that returns the API version.

src/bindings.d.ts
fforbeck: I think it will be better if we use separate files for each middleware/handler instead of defining everything in `middleware.js`, otherwise it will be hard to maintain all of them in one place. Then I can refactor this as well.

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
  * @typedef {import('@web3-storage/gateway-lib').UnixfsContext} UnixfsContext
  */
 
+/**
+ *
+ * @param {string} s
+ * @returns {import('./bindings.js').TokenMetadata}
+ */
+function deserializeTokenMetadata (s) {
```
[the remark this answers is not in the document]
fforbeck: Agree. We already have the `handlers` directory, I will use that to define the rate-limit.js middleware.

src/middleware.js
fforbeck: The issue I see with passing the cache and accounting to this function is that we will always need to call the `Accounting.create...` to provide that instance in the function call, and if the value is in the cache already we don't even need to execute that call at all because it just decodes the cached value.

src/middleware.js
fforbeck: Good catch. Yes, I believe we will need to use the `ctx.waitUntil` call here.

src/middleware.js
fforbeck: It is typed as a boolean in the Environment bindings.d.ts

src/middleware.js
```diff
@@ -94,7 +247,8 @@ export function withContentClaimsDagula (handler) {
 export function withVersionHeader (handler) {
   return async (request, env, ctx) => {
     const response = await handler(request, env, ctx)
-    response.headers.set('x-freeway-version', version)
+    // @ts-expect-error The type definition for env.VERSION is defined but it is not detected by typescript
+    response.headers.set('x-freeway-version', env.VERSION || 'unknown')
```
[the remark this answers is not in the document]
fforbeck: My bad. Fixed the type declaration and now we don't even need that conditional.

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+/**
+ *
+ * @param {Environment} env
+ * @param {string} authToken
+ * @returns TokenMetadata
+ */
+async function getTokenMetadata (env, authToken) {
+  const cachedValue = await env.AUTH_TOKEN_METADATA.get(authToken)
+  // TODO: we should implement an SWR pattern here - record an expiry in the metadata and if the expiry has passed, re-validate the cache after
+  // returning the value
```
Peeja, line 67: **suggestion:** Let's not leave this as a `TODO` here; let's capture the user value and move it to a card we can prioritize on the board. We can't decide to actually do this until it's in the flow and context of our other work.

src/middleware.js
Peeja, line 74: **suggestion:** Should we add an `expirationTtl` on this? Right now (especially given we're not SWRing), we appear to cache these indefinitely.

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+          // this means token based requests will be subject to normal rate limits until the data propagates
+          return checkRateLimitForCID(env, cid)
+        }
+      } else {
+        // no token, use normal rate limit
+        return checkRateLimitForCID(env, cid)
+      }
+    }
+  })
+}
```
Peeja, line 113: **suggestion:** Untangle this a bit so we have the one path to `RATE_LIMIT_EXCEEDED.NO` followed by one fallback to `checkRateLimitForCID()`.

```suggestion
const RateLimits = {
  create: ({ env }) => ({
    check: async (cid, options) => {
      // We'll serve the request without a rate limit if...
      const authToken = await getAuthorizationTokenFromRequest(options)

      // ...there's an auth token in the request...
      if (authToken) {
        console.log(`found token ${authToken}, looking for content commitment`)
        const tokenMetadata = await getTokenMetadata(env, authToken)

        // ...and we can find a content commitment for the token...
        if (tokenMetadata) {
          // ...and the token is valid.
          if (!tokenMetadata.invalid) {
            return RATE_LIMIT_EXCEEDED.NO
          }
        }
      }

      // Otherwise, we'll rate limit the request.
      return checkRateLimitForCID(env, cid)
    }
  })
}
```

src/bindings.d.ts
```diff
@@ -1,7 +1,47 @@
...
+  RATE_LIMITER: RateLimit
+  AUTH_TOKEN_METADATA: KVNamespace
+  FF_RATE_LIMITER_ENABLED: boolean
+}
+
+export type GetCIDRequestData = Pick<Request, 'url' | 'headers'>
+
+export type GetCIDRequestOptions = GetCIDRequestData
+
+export interface RateLimitsService {
```
Peeja, line 26: **suggestion:** This type didn't exist yet, so it was `any`.

```suggestion
export type RateLimitExceeded = typeof RATE_LIMIT_EXCEEDED[keyof typeof RATE_LIMIT_EXCEEDED]

```

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+    return RATE_LIMIT_EXCEEDED.YES
+  }
+}
+
+/**
+ *
+ * @param {Environment} env
+ * @param {string} authToken
+ * @returns TokenMetadata
+ */
```
Peeja, line 63: **suggestion:** Actually type this. 😜

```suggestion
 * @returns {Promise<import('./bindings.js').TokenMetadata | null>}
```

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+      return handler(request, env, ctx)
+    }
+
+    const { dataCid } = ctx
+
+    const rateLimits = RateLimits.create({ env })
+    const isRateLimitExceeded = await rateLimits.check(dataCid, request)
+
+    if (isRateLimitExceeded === RATE_LIMIT_EXCEEDED.YES) {
+      // TODO should we record this?
```
Peeja, line 160: **suggestion:** I think we don't need to—we do get some basic HTTP logging for free on Cloudflare, right?

Peeja (changes requested):

Peeja:
FWIW, I'm highly in favor of Jest. It's not legacy; quite the opposite, AFAIK it's where the community primarily is right now. Granted I'm biased by being used to Jest, but I'd love to decide to (in the long run) move _everything_ to Jest.

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+ */
+function serializeTokenMetadata (m) {
+  // TODO should this be dag-json?
+  return JSON.stringify(m)
+}
+
+/**
+ *
+ * @param {Environment} env
+ * @param {import('@web3-storage/gateway-lib/handlers').CID} cid
```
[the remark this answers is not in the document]
fforbeck: @alanshaw the issue is that `IpfsUrlContext` returns the `dataCid` as type `multiformats/cid`. Should we use the same?

wrangler.toml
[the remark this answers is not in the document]
fforbeck: Yes, that's right!

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+}
+
+/**
+ *
+ * @param {Pick<Request, 'headers'>} request
+ * @returns string
+ */
+async function getAuthorizationTokenFromRequest (request) {
+  // TODO this is probably wrong
+  const authToken = request.headers.get('Authorization')
```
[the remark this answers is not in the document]
fforbeck: Makes sense. I will change that. Thanks

README.md
```diff
@@ -10,6 +10,80 @@
...
+simple = { limit = 100, period = 60 }
+
+[[env.YOUR_USERNAME.kv_namespaces]]
+binding = "AUTH_TOKEN_METADATA"
+id = "YOUR_TOKEN" //FIXME how to obtain this token?
+```
+
+5. Start local server
+```sh
+npx wrangler dev -e YOUR_USERNAME
```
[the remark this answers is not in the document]
fforbeck: I think having them in the Readme should be fine for now. That `dev` script starts the worker in dev mode (but that was for miniflare v2), I've updated it to miniflare v3 and I guess we can remove that `dev` script for now because they recommend using `npx wrangler dev -e <profile>` instead.

.github/actions/test/action.yml
```diff
@@ -13,5 +13,9 @@ runs:
       shell: bash
     - run: npm run lint
       shell: bash
-    - run: npm test
+    - run: npm run test:unit
+      name: Unit Tests
+      shell: bash
+    - run: npm run test
+      name: Miniflare Integration Tests
```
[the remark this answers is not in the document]
fforbeck: Yeah. Good point. I will change that.

src/middleware.js
fforbeck: Thanks for the suggestion. I leave this to be changed in another PR.

src/middleware.js
alanshaw: Hmm, yes I see but that doesn't automatically turn it into a boolean - it's just an environment variable right, either set in `wrangler.toml` or as a deployment variable - it will be a string unless you explicitly cast it in code.

set-version-env.js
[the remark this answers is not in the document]
alanshaw: Oh ok got it - because new tests are running against actual source code not esbuild bundle.

Please next time just leave a comment in the PR description to explain! 🙏

src/middleware.js
```diff
@@ -15,6 +14,160 @@ import { handleCarBlock } from './handlers/car-block.js'
...
+/**
+ *
+ * @param {Environment} env
+ * @param {import('@web3-storage/gateway-lib/handlers').CID} cid
+ * @returns {Promise<import('./constants.js').RATE_LIMIT_EXCEEDED>}
+ */
+async function checkRateLimitForCID (env, cid) {
+  const rateLimiter = env.RATE_LIMITER
+  if (!rateLimiter) {
+    console.warn('no rate limiter found')
```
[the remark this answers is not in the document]
alanshaw: ACK yes agree. We should throw an error if rate limter is not configured then.

src/middleware.js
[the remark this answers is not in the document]
alanshaw: Oh, yes I guess it is old, I'd use `import('multiformats').CID` instead here then.

src/handlers/rate-limiter.js
```diff
@@ -0,0 +1,151 @@
...
+ * @param {string} authToken
+ * @param {import('@web3-storage/gateway-lib').Context} ctx
+ * @returns {Promise<import('../bindings.js').TokenMetadata | null>}
+ */
+async function getTokenMetadata (env, authToken, ctx) {
+  const cachedValue = await env.AUTH_TOKEN_METADATA.get(authToken)
+  // TODO: we should implement an SWR pattern here - record an expiry in the metadata and if the expiry has passed, re-validate the cache after
+  // returning the value
+  if (cachedValue) {
+    return decode(cachedValue)
```
alanshaw, line 121: You can `get` and `put` and JSON decode/dnecode so you perhaps don't need these encode/decode functions?:

https://developers.cloudflare.com/kv/api/read-key-value-pairs/#get-method

alanshaw (changes requested):

src/middleware.js
alanshaw: Yes, understood but I'm not convinced it'll be a performance problem.

Related, I'd probably implement the cache as a transparent wrapper around the accounting service and pass the accounting service as a dependency to the rate limiting service.

i.e. something like:

```js
const accounting = Accounting.create({ serviceURL: env.ACCOUNTING_SERVICE_URL })
const cachedAccounting = Accounting.withCache(accounting, env.AUTH_TOKEN_METADATA)
const rateLimiter = RateLimiter.create(cachedAccounting)

const status = await rateLimiter.check(cid, request)
if (status  === RATE_LIMIT_EXCEEDED.YES) {
  // ...
```

So then it should be easier to test the rate limiter, since the rate limiter is not importing a function that requires a KV _and_ an accounting service. Also easier to test the caching logic in isolation.

...anyways...just a suggestion - non blocking.

src/middleware.js
fforbeck: Oh ok. That makes sense. Thanks for the clarification.

src/handlers/rate-limiter.js
fforbeck: @alanshaw The current API doesn't support that `json` type. I won't change this now because this PR is getting too big, but will certainly update the cloudflare packages in another PR to use the latest version available - which supports `json` values.

src/handlers/rate-limiter.js
```diff
@@ -0,0 +1,151 @@
...
+  }
+}
+
+/**
+ * @param {import("../bindings.js").Environment} env
+ * @param {string} authToken
+ * @param {import('@web3-storage/gateway-lib').Context} ctx
+ * @returns {Promise<import('../bindings.js').TokenMetadata | null>}
+ */
+async function getTokenMetadata (env, authToken, ctx) {
```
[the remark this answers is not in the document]
fforbeck: We do use that to call `ctx.waitUntil` at [line 128]( https://github.com/storacha/freeway/pull/115/files/e37c45c88caebca3f35314cb9d8a5b98c303baf5#diff-29338ef628c3debc8fcd22bcfa9b84b082aa8be54c8c372bb071537b234159c3R128).

fforbeck:
@alanshaw I've applied pretty much all the suggestions, however, I left some of them to be done in different PRs to minimize the number of changes here.
TODO
- Rename URLs from Readme
- Implement the cache as a transparent wrapper around the accounting service and pass the accounting service as a dependency on the rate-limiting service. https://github.com/storacha/freeway/pull/115#discussion_r1784309869
- Log "rate limit disabled" just once at startup please not every request
- Update Cloudflare workers library to use the latest KV `put/get` types - which support JSON values

src/bindings.d.ts
```diff
@@ -1,7 +1,38 @@
...
+  invalid?: boolean
 }
+
+export interface AccountingService {
+  record: (cid: CID, options: GetCIDRequestConfig) => Promise<void>
+  getTokenMetadata: (token: string) => Promise<TokenMetadata | null>
+}
+
+export interface Accounting {
+  create: ({ serviceURL }: { serviceURL?: string }) => AccountingService
```
alanshaw, line 36: Minor, but the same with this one - we'd typically not define this in types and just define it in the file and use like:

```js
import * as Accounting from 'services/accounting.js'

Accounting.create(...)
```

...and in `accounting.js`

```js
export const create = () => { /* ... */ }
```

alanshaw (approved):

src/bindings.d.ts
fforbeck: Sounds good. I will add it to the TODO list as well. Thanks

travis (approved):

PR #119, opened 2024-10-08
refactor: Type check and restructure tests

Peeja:
This is more what I was getting at with https://github.com/storacha/freeway/pull/116/files#r1787752932. This avoids some of the stashing and mutation by putting more inside the `new Promise()` construction itself.

I've also pulled `wranglerEnv` out into its own `const`, because it's actually an _input_, not an output. Then I've made `workerInfo` a `let` which is assigned when Wrangler boots, and is `undefined` before that, signaling that Wrangler is not yet available. The previous code was failing because there were `undefined` fields that aren't typed to allow `undefined`. Which wasn't an issue until…

I also turned type checking on in `test/`, because apparently it wasn't already. 😛 And lastly, I've ignored `.wrangler`, because Wrangler writes to that when it runs.

### DIFF
--- a/test/helpers/run-wrangler.js
+++ b/test/helpers/run-wrangler.js
@@ -103,7 +103,7 @@ async function runLongLivedWrangler (
           wranglerProcess.pid,
                   `Command "${command.join(' ')}" had no process id`
         )
-        treeKill(wranglerProcess.pid, (e) => {
+        treeKill(wranglerProcess.pid, 'SIGKILL', (e) => {
           if (e) {
             console.error(
               'Failed to kill command: ' + command.join(' '),
--- a/test/helpers/run-wrangler.js
+++ b/test/helpers/run-wrangler.js
@@ -62,70 +62,58 @@ async function runLongLivedWrangler (
   cwd,
   env
 ) {
-  let settledReadyPromise = false
-  /** @type {(value: { ip: string port: number }) => void} */
-  let resolveReadyPromise
-  /** @type {(reason: unknown) => void} */
-  let rejectReadyPromise
-
-  const ready = new Promise((resolve, reject) => {
-    resolveReadyPromise = resolve
-    rejectReadyPromise = reject
-  })
-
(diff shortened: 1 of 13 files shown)


fforbeck (changes requested):
That's awesome! Thanks for updating that @Peeja.
I guess you just need to sort out those linter/type issues flagged by the test workflow, and it is good to go. I also tested this locally and worked well.

fforbeck:
@Peeja For some reason after the changes the `workerd` (wrangler) process is not terminated when the integration tests are completed. I can see it running in the background after the test suite is completed.

test/helpers/run-wrangler.js
```diff
@@ -62,70 +62,58 @@ async function runLongLivedWrangler (
...
-        }
-        resolve()
+    /** @type {WranglerProcessInfo['stop']} */
+    async function stop () {
+      return new Promise((resolve) => {
+        assert(
+          wranglerProcess.pid,
+                  `Command "${command.join(' ')}" had no process id`
+        )
+        treeKill(wranglerProcess.pid, (e) => {
```
fforbeck, line 106: **suggestion:**
I've tested this locally, and `SIGKILL` was the only reliable way to fully terminate the process after the test was completed. The default `SIGTERM` (standard termination signal) left the process running in the background, likely because the process did not handle the signal as expected.

Since `SIGKILL` forcefully terminates the process without allowing for cleanup, it's a more immediate solution that ensures processes don't linger in the background. This is especially useful for our test environment. While this may bypass graceful shutdown procedures, it prevents the accumulation of stuck processes when running tests multiple times, ensuring a clean state in development.

```suggestion
        treeKill(wranglerProcess.pid, 'SIGKILL', (e) => {
```

fforbeck (approved):
This looks great to me! However, since the changes affect more than just the Wrangler setup, we should consider renaming the PR to reflect its broader scope. 

I also added a minor suggestion about the `treeKill` usage to kill the wrangler process.

PR #120, opened 2024-10-28
feat: egress tracker middleware

fforbeck:
### Egress Tracker Middleware

**Summary:**
This PR introduces an egress tracking middleware for the Freeway project, enabling accurate measurement and recording of egress data for served content. The middleware tracks the bytes sent in each response body and logs them with the accounting service tied to each content ID (CID).

**Key Changes:**
- **Egress Middleware (`withEgressHandler`)**: 
  - Wraps response handlers to track and count bytes sent to the client.
  - Controlled by the `FF_EGRESS_TRACKER_ENABLED` feature flag, enabling or disabling egress tracking as needed. It is disabled by default.

- **Accounting Service Integration**:
  - Logs egress data with the accounting service, using either an `ACCOUNTING_SERVICE` from the context or a new instance based on the `ACCOUNTING_SERVICE_URL` environment variable.
  - Egress data is linked to the CID of the served content, ensuring precise tracking. (The actual accounting service implementation, integrating `w3up-client` for the new `usage/record` capability, will follow in a separate PR.)

- **Efficient Byte Counting via `TransformStream`**:
  - Utilizes a `TransformStream` (`createEgressPassThroughStream`) to passively count bytes in the response body without altering content.
  - On stream completion, the `flush` method records total egress to the accounting service using `ctx.waitUntil()` for non-blocking calls.

**Error Handling**:
- Logs errors encountered during data streaming and halts byte counting without interrupting the original response chain. This ensures resilience even in cases of partial or interrupted streams.

**Testing**:
- Added thorough tests to validate egress recording across scenarios, including complete responses, interrupted streams, and error cases.

**Next Steps**:
- Integration tests for verifying egress tracking accuracy and accounting service interactions in various streaming conditions (planned for a future PR).
- `w3up-client` integration to execute the new `usage/record` capability in subsequent development.


src/services/accounting.js
```diff
@@ -3,8 +3,8 @@
  */
 export const Accounting = {
   create: ({ serviceURL }) => ({
-    record: async (cid, options) => {
-      console.log(`using ${serviceURL} to record a GET for ${cid} with options`, options)
+    record: async (cid, bytes, servedAt) => {
```
fforbeck, line 6: I will create another PR to integrate with `w3up-client` and execute the `usage/record` invocation from here.

test/unit/middlewares/egress-tracker.spec.js
```diff
@@ -0,0 +1,462 @@
...
+
+      /** @type {(import('carstream').Block & import('carstream').Position)[]} */
+      const blocks = []
+      await source
+        .pipeThrough(new CARReaderStream())
+        .pipeTo(new WritableStream({
+          write: (block) => { blocks.push(block) }
+        }))
+
+      // expect(blocks[0].bytes).to.deep.equal(carBytes) - FIXME (fforbeck): how to get the correct byte count?
```
fforbeck, line 217: Still need to figure out how to check the byte count of the returned CAR file.

src/handlers/egress-tracker.js
```diff
@@ -0,0 +1,117 @@
...
+     * It is called only once and it triggers a non-blocking call to the accounting service.
+     * If an error occurs, the egress is not recorded.
+     * NOTE: The flush function is NOT called in case of an stream error.
+     */
+    async flush (controller) {
+      try {
+        // Non-blocking call to the accounting service to record egress
+        if (totalBytesServed > 0) {
+          ctx.waitUntil(accounting.record(dataCid, totalBytesServed, new Date().toISOString()))
+        }
```
Peeja, line 110: **thought:** I notice that:

* `dataCid` and `accounting` are arguments that we were passed.
* `ctx.waitUntil` is the only thing we use from the `ctx`.
* Since anyone can `new Date().toISOString()`, that leaves `totalBytesServed` (understandably) as the only thing that this function is really responsible for creating.
* It's a little odd that `createEgressPassThroughStream` knows how to communicate with the Accounting Service.

**suggestion:** What about something like:

```ts
/**
 * @param {(totalBytesServed: number) => void} onClose
 * @returns {TransformStream}
 */
function createByteCountStream(onClose) {
  let totalBytesServed = 0;

  return new TransformStream({
    // ...
    async flush() {
      onClose(totalBytesServed);
    },
  });
}

const responseBody = response.body.pipeThrough(
  createByteCountStream((totalBytesServed) => {
    // Non-blocking call to the accounting service to record egress
    if (totalBytesServed > 0) {
      ctx.waitUntil(
        accounting.record(dataCid, totalBytesServed, new Date().toISOString())
      );
    }
  })
);
```

src/handlers/egress-tracker.js
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/**
+ * Creates a TransformStream to count bytes served to the client.
+ * It records egress when the stream is finalized without an error.
+ *
+ * @param {import('@web3-storage/gateway-lib/middleware').Context} ctx - The context object.
+ * @param {AccountingService} accounting - The accounting service instance to record egress.
+ * @param {import('@web3-storage/gateway-lib/handlers').CID} dataCid - The CID of the served content.
+ * @returns {TransformStream} - The created TransformStream.
```
Peeja, line 66: **suggestion:** Let's fully type this:

```ts
/**
 * @template {Uint8Array} T
 * @returns {TransformStream<T, T>} - The created TransformStream.
 */
```

That *should* also mean that you get to drop the explicit `@param` annotations inside the function, as the types should flow from there.

src/bindings.d.ts
```diff
@@ -1,18 +1,20 @@
...
-export interface Environment extends CarBlockEnvironment, RateLimiterEnvironment {
+export interface Environment extends CarBlockEnvironment, RateLimiterEnvironment, EgressTrackerEnvironment {
   VERSION: string
   CONTENT_CLAIMS_SERVICE_URL?: string
+  ACCOUNTING_SERVICE_URL: string
 }
 
 export interface AccountingService {
-  record: (cid: CID, options: GetCIDRequestConfig) => Promise<void>
+  record: (resource: UnknownLink, bytes: number, servedAt: string) => Promise<void>
```
Peeja, line 13: **thought:** I think we need the Space here, not the CID: the CID doesn't correspond to (necessarily) exactly one Space, and therefore customer. Luckily, we'll have that in the context with my PR!

src/handlers/rate-limiter.js
```diff
@@ -127,9 +122,7 @@ async function getTokenMetadata (env, authToken, ctx) {
     return decode(cachedValue)
   }
 
-  const accounting = Accounting.create({
-    serviceURL: env.ACCOUNTING_SERVICE_URL
-  })
+  const accounting = ctx.ACCOUNTING_SERVICE ?? Accounting.create({ serviceURL: env.ACCOUNTING_SERVICE_URL })
```
Peeja, line 125: **suggestion:** Another reason to always create the `ACCOUNTING_SERVICE` higher up the stack: we don't need to duplicate the logic to create it in each place we use it.

test/unit/middlewares/egress-tracker.spec.js
```diff
@@ -0,0 +1,462 @@
...
+    ACCOUNTING_SERVICE_URL: 'http://example.com',
+    FF_EGRESS_TRACKER_ENABLED: 'true'
+  })
+
+const accountingRecordMethodStub = sinon.stub()
+  // @ts-expect-error
+  .returns(async (cid, bytes, servedAt) => {
+    console.log(`[mock] record called with cid: ${cid}, bytes: ${bytes}, servedAt: ${servedAt}`)
+    return Promise.resolve()
+  })
```
Peeja, line 42: **suggestion:** To lose this `@ts-expect-error`:

```suggestion
const accountingRecordMethodStub = sinon.stub().returns(
  /** @type {import('../../../src/bindings.js').AccountingService['record']} */
  async (cid, bytes, servedAt) => {
    console.log(
      `[mock] record called with cid: ${cid}, bytes: ${bytes}, servedAt: ${servedAt}`
    );
    return Promise.resolve();
  }
);
```

test/unit/middlewares/egress-tracker.spec.js
```diff
@@ -0,0 +1,462 @@
...
+    DEBUG: 'true',
+    ACCOUNTING_SERVICE_URL: 'http://example.com',
+    FF_EGRESS_TRACKER_ENABLED: 'true'
+  })
+
+const accountingRecordMethodStub = sinon.stub()
+  // @ts-expect-error
+  .returns(async (cid, bytes, servedAt) => {
+    console.log(`[mock] record called with cid: ${cid}, bytes: ${bytes}, servedAt: ${servedAt}`)
+    return Promise.resolve()
```
Peeja, line 41: **suggestion:** If I'm reading right, we can just drop this line altogether. It's the same as just `return`.

Peeja (changes requested):
Looking good! Some suggestions:

test/unit/middlewares/egress-tracker.spec.js
hannahhoward: Async funcs automatically convert return values to promises, so yes.

src/handlers/egress-tracker.js
hannahhoward: reasonable

hannahhoward: will certainly simplify the test

hannahhoward (approved):
LGTM but agree with @Peeja 's comments

src/handlers/egress-tracker.js
```diff
@@ -0,0 +1,117 @@
...
+
+    const { readable, writable } = createEgressPassThroughStream(ctx, accounting, dataCid)
+
+    try {
+      ctx.waitUntil(response.body.pipeTo(writable))
+    } catch (error) {
+      console.error('Error in egress tracker handler:', error)
+      // Original response in case of an error to avoid breaking the chain and serve the content
+      return response
+    }
```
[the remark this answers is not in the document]
fforbeck: @Peeja - woops, it won't catch. It needs to be a .catch in the `pipeTo` promise. I will fix that. Thanks

src/handlers/egress-tracker.js
```diff
@@ -0,0 +1,117 @@
...
+function createEgressPassThroughStream (ctx, accounting, dataCid) {
+  let totalBytesServed = 0
+
+  return new TransformStream({
+    /**
+     * The start function is called when the stream is being initialized.
+     * It resets the total bytes served to 0.
+     */
+    start () {
+      totalBytesServed = 0
```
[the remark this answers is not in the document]
fforbeck: Yeah, we don't need to set it twice. I will remove the `start` function.

src/handlers/egress-tracker.js
```diff
@@ -0,0 +1,117 @@
...
+    async flush (controller) {
+      try {
+        // Non-blocking call to the accounting service to record egress
+        if (totalBytesServed > 0) {
+          ctx.waitUntil(accounting.record(dataCid, totalBytesServed, new Date().toISOString()))
+        }
+      } catch (error) {
+        console.error('Error while recording egress:', error)
+        controller.error(error)
+      }
```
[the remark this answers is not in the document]
fforbeck: We could capture the error with something like:
```js
// Non-blocking call to the accounting service to record egress
      if (totalBytesServed > 0) {
        const recordPromise = accounting
          .record(dataCid, totalBytesServed, new Date().toISOString())
          .catch(error => {
            console.error('Error while recording egress:', error)
            controller.error(error)
          })
        ctx.waitUntil(recordPromise)
      }
```

src/bindings.d.ts
fforbeck: That's right. I will add that one in the PR that integrates with `ucanto` lib.

test/unit/middlewares/egress-tracker.spec.js
fforbeck: @Peeja / @hannahhoward, any advice on comparing the total bytes of the returned CAR file with `expectedTotalBytes`? I suspect I'm not converting the file correctly. Any tips would be helpful!

src/handlers/egress-tracker.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    if (!response.ok || !response.body) {
+      return response
+    }
+
+    const { dataCid } = ctx
+    const accounting = ctx.ACCOUNTING_SERVICE ?? Accounting.create({
+      serviceURL: env.ACCOUNTING_SERVICE_URL
+    })
```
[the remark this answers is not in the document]
fforbeck: Agree. I will add that in the new PR too.

Peeja (changes requested):
Woohoo! Looking even better!

I dug into the tests a bit more. My main concern is that they're a bit hard to follow when you don't have all the context loaded in your brain. Hopefully these suggestions improve that?

fforbeck:
@Peeja - thanks for the review! Since these suggestions are test-related and don’t seem to be blockers, would you be open to merging this now and handling the remaining refactor in [PR #122](https://github.com/storacha/freeway/pull/122)? (which will affect the egress handler and accounting service anyway).

src/middleware/withEgressTracker.js
```diff
@@ -0,0 +1,92 @@
+import { Accounting } from '../services/accounting.js'
+
+/**
+ * @import { Context, IpfsUrlContext, Middleware } from '@web3-storage/gateway-lib'
+ * @import { Environment } from './withEgressTracker.types.js'
+ * @import { AccountingService } from '../bindings.js'
+ * @typedef {IpfsUrlContext & { ACCOUNTING_SERVICE?: AccountingService }} EgressTrackerContext
```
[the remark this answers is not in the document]
fforbeck: Yep, I've changed that already in another PR: https://github.com/storacha/freeway/pull/122/files#diff-3a7afa3c27e9d98e1640d32665e7e31f94de5e048ff9d7f138b2e5f068300ee3R11

Peeja (approved):
Approving so test changes can happen in https://github.com/storacha/freeway/pull/122

test/unit/middleware/withEgressTracker.spec.js
```diff
@@ -0,0 +1,462 @@
...
+      const response = await handler(request, env, ctx)
+      // Ensure the response body is fully consumed
+      const responseBody = await response.text()
+
+      expect(response.status).to.equal(200)
+      expect(responseBody).to.equal('Hello, world!')
+      expect(accountingRecordMethodStub.calledOnce, 'record should be called once').to.be.true
+      expect(accountingRecordMethodStub.args[0][0], 'first argument should be the cid').to.equal(ctx.dataCid)
+      expect(accountingRecordMethodStub.args[0][1], 'second argument should be the total bytes').to.equal(totalBytes)
+    }).timeout(10_000)
```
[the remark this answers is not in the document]
fforbeck: We don't. It came from other test cases. I will clean that up.

test/unit/middleware/withEgressTracker.spec.js
```diff
@@ -0,0 +1,462 @@
...
+      const handler = withEgressTracker(innerHandler)
+      const request = await createRequest()
+
+      const response = await handler(request, env, ctx)
+      const responseBody = await response.text()
+
+      expect(response.status).to.equal(200)
+      expect(responseBody).to.equal('Delayed response content')
+      expect(accountingRecordMethodStub.calledOnce, 'record should be called once').to.be.true
+      expect(accountingRecordMethodStub.args[0][1], 'second argument should be the total bytes').to.equal(totalBytes)
```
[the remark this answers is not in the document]
fforbeck: I think this breaks the flow a bit. Most tests follow the same structure, making it easy to skim through, but this one stands out and requires more detailed attention.

test/unit/middleware/withEgressTracker.spec.js
```diff
@@ -0,0 +1,462 @@
...
+
+      const response = await handler(request, env, ctx)
+
+      expect(response.status).to.equal(200)
+      expect(accountingRecordMethodStub.called, 'record should not be called').to.be.false
+    })
+  })
+
+  describe('withEgressTracker -> Concurrent Requests', () => {
+    it('should correctly track egress for multiple concurrent requests', async () => {
```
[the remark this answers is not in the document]
fforbeck: Even if the requests don't really interact with each other, this test can still be useful. It acts like a safety check to make sure the system handles multiple requests smoothly without any unexpected hiccups.

PR #122, opened 2024-10-30
feat: accounting service

fforbeck:


fforbeck:
moved to https://github.com/storacha/freeway/pull/123

PR #123, opened 2024-10-31
feat: egress client - ucanto integration

fforbeck:
### Context

This PR introduces `EgressClient` to the application context, enabling capability invocations such as `Space.egressRecord` for logging egress directly to the UCanto Server (Upload API). With this integration, the previous Accounting Service is now replaced by a method in `EgressClient` for egress logging.

### Key Changes

- **Middleware Integration**: Added `withEgressClient` middleware to simplify capability invocations across the application.
  
- **EgressClient Creation**: Introduced a `create` function to instantiate `UCantoClient`, establishing a secure connection to the UCanto Server based on environment configurations.
  
- **Egress Recording**: Refactored `EgressClient.record` function to use the `Space.egressRecord` capability, allowing for efficient egress byte tracking in our infrastructure.
  
- **Connection Management**: Added a `connect` function to handle connection setup with the UCanto Server.

- **Environment Variables**: Updated the environment variables, including service Web DIDs and service URLs.

- **Enhanced Context**: Added the `delegationProofs` to the application context, so we can use that information to invoke the `egressRecord` capabilities. Also added the `GatewayIdentity` to the application context.

- **Accounting Service**: There is no Accounting Service anymore. Instead, we use the EgressClient to record the egress event.

- **Wrangler**: Updated the configs for all environments and the wrangler lib to the latest version.

- **Telemetry**: Added a feature flag for Open Telemetry - if enabled, the Egress Record call will fail to execute the `this.fetch` function call. See https://github.com/storacha/project-tracking/issues/176 for more details.


src/middleware/withUcantoClient.js
```diff
@@ -0,0 +1,109 @@
...
+ * @param {import('@ucanto/principal/ed25519').DIDKey} space - The Space DID where the content was served
+ * @param {import('@ucanto/principal/ed25519').UnknownLink} resource - The link to the resource that was served
+ * @param {number} bytes - The number of bytes served
+ * @param {Date} servedAt - The timestamp of when the content was served
+ * @param {import('@ucanto/principal/ed25519').EdSigner} principal - The principal signer
+ * @param {Signer.Verifier} service - The service verifier
+ * @param {any} connection - The connection to execute the command
+ * @returns {Promise<void>}
+ */
+async function recordEgress (space, resource, bytes, servedAt, principal, service, connection) {
```
Peeja, line 94: **suggestion (non-blocking):** I think I'd either keep this to being a pure Ucanto client (which I guess is properly termed a Ucanto "connection") and move the invocation into `withRateLimit`, or change the thing in the context from `ucantoClient` to…maybe `usageClient`? A "Ucanto client" wouldn't know about our specific (cap)abilities.

I think the latter makes a lot of sense given the way we've defined "clients" in `w3up`, and it's also the easiest change. 😄

Peeja (approved):
🫵🥫➡️❗

Looks great! One suggestion:

travis (approved):
lgtm! agree with Petra but happy to see this ship with or without the change

src/middleware/withAuthorizedSpace.js
```diff
@@ -120,14 +122,22 @@ const authorize = async (space, ctx) => {
     capability: serve.transportHttp,
     authority: ctx.gatewayIdentity,
     principal: Verifier,
-    validateAuthorization: () => ok({})
+    validateAuthorization: () => ok({}),
+    // resolveDIDKey: async (did) => {
+    //   if (did === ctx.gatewayIdentity.did()) return ok(ctx.gatewayIdentity.toDIDKey())
+    //   throw new Error(`Unknown DID: ${did}`)
+    // }
```
Peeja, line 129: **suggestion:** Obviously, let's not have commented lines here, but…this makes me wonder if this is actually the correct way to do this after all, rather than a `Signer` with a swapped-in DID. The _correct_ correct way to do it is to actually resolve the DID document and get the key, but we don't currently publish DID documents for our `did:web:`s, so we're not there yet.

In any case, the thing to do right now is whatever's easiest, but maybe using `resolveDIDKey` makes sense? 🤔

src/middleware/withAuthorizedSpace.types.ts
```diff
@@ -1,10 +1,12 @@
...
+import { GatewayIdentityContext as GatewayIdentityContext } from './withGatewayIdentity.types.js'
 
-export interface DelegationsStorageContext extends MiddlewareContext {
-  /** The identity of the gateway itself */
-  gatewayIdentity: Ucanto.Signer
+export interface DelegationsStorageContext
+  extends MiddlewareContext,
+    GatewayIdentityContext {
   delegationsStorage: DelegationsStorage
+  delegationProofs?: Ucanto.Delegation[]
```
Peeja, line 9: **suggestion:** Would you add some JSDoc for `delegationProofs`? It's not obvious from the name what we should expect to find in there.

src/middleware/withAuthorizedSpace.js
```diff
@@ -57,17 +58,18 @@ export function withAuthorizedSpace (handler) {
 
     try {
       // First space to successfully authorize is the one we'll use.
-      const space = await Promise.any(
+      const { space: selectedSpace, delegationProofs } = await Promise.any(
```
Peeja, line 61: **praise:** Love the way `delegationProofs` come back through now!

src/middleware/withAuthorizedSpace.types.ts
```diff
@@ -1,10 +1,12 @@
 import * as Ucanto from '@ucanto/interface'
 import { Context as MiddlewareContext } from '@web3-storage/gateway-lib'
+import { GatewayIdentityContext as GatewayIdentityContext } from './withGatewayIdentity.types.js'
 
-export interface DelegationsStorageContext extends MiddlewareContext {
-  /** The identity of the gateway itself */
-  gatewayIdentity: Ucanto.Signer
+export interface DelegationsStorageContext
+  extends MiddlewareContext,
+    GatewayIdentityContext {
```
Peeja, line 7: **suggestion:** Rather than include `GatewayIdentityContext` in this, let's keep them separate and just add it to the intersection of context types coming out of the middleware. That way they're each their own thing.

src/middleware/withEgressClient.js
```diff
@@ -0,0 +1,101 @@
...
+  const invocation = invoke({
+    issuer: ctx.gatewayIdentity,
+    audience: ctx.gatewayIdentity,
+    with: SpaceDID.from(space),
+    nb: {
+      resource,
+      bytes,
+      servedAt: Math.floor(servedAt.getTime() / 1000)
+    },
+    proofs: ctx.delegationProofs ? ctx.delegationProofs : []
```
Peeja, line 94: **question:** _Can_ this be missing? Don't we know by type that it has to be present?

src/middleware/withEgressClient.js
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/**
+ * Records the egress bytes in the UCanto Server by invoking the `Space.egressRecord` capability.
+ *
+ * @param {import('@ucanto/principal/ed25519').DIDKey} space - The Space DID where the content was served
+ * @param {import('@ucanto/principal/ed25519').UnknownLink} resource - The link to the resource that was served
+ * @param {number} bytes - The number of bytes served
+ * @param {Date} servedAt - The timestamp of when the content was served
+ * @param {any} connection - The connection to execute the command
```
Peeja, line 77: **suggestion:** Can we type this for real?

test/helpers/builder.js
```diff
@@ -19,7 +19,9 @@ export class Builder {
    * @returns {Promise<{ root: import('multiformats').UnknownLink, shards: import('multiformats').Link[]}>}
    */
   async add (input, options = {}) {
-    console.log('Adding ' + (Array.isArray(input) ? `${input.length} file${input.length === 1 ? '' : 's'}` : '1 blob') + '...')
+    if (process.env.DEBUG) {
```
Peeja, line 22: **praise:** 👏🏻

Peeja (commented):
Looking good! Suggestions and thoughts below. Looks like this needs some type love and some `npm run lint:fix`ing too.

src/middleware/withUcantoClient.js
fforbeck: Thanks. I've renamed that to `egressClient`.

src/middleware/withEgressClient.js
fforbeck: It can't. I've fixed that.

PR #125, opened 2024-11-01
refactor(handler): at opentelemtry instrumentation

hannahhoward:
# Goals

Gain insight into what's going on with freeway

# Implementation

When in doubt, instrument. This adds opentelemtry and a default cloudflare open telemetry implementation, that at minimum tells you how long fetch requests are taking. And it's.... very interesting...

Screenshot 2024-11-01 at 1 35 23 AM

This is a trace of about a 64mb file downloading. 

I dunno what to make of all it, except to say... ouch those CARPARK downloads take a WHILE. I wonder if somehow accessing them through public URLs is a lot slower?


package.json
```diff
@@ -69,5 +72,6 @@
     "ignore": [
       "*.ts"
     ]
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
alanshaw, line 76: You're using yarn?

src/index.js
```diff
@@ -76,6 +78,29 @@ export default {
...
+/**
+ *
+ * @param {Environment} env
+ * @param {*} _trigger
+ */
+function config (env, _trigger) {
+  if (env.HONEYCOMB_API_KEY) {
+    return {
+      exporter: {
+        url: 'https://api.honeycomb.io/v1/traces',
```
alanshaw, line 90: Does it get sent for every request? I'm slightly worried we'll run up a large bill or get rate limited immediately....

alanshaw (approved):
I don't really understand it, but I like the screenshot and it looks like it'll give us some good insights.

I would remove the yarn package manager thing from `package.json` unless it's required...

alanshaw:
Great insights. 

Also, what the heck is happening between the carpark fetches?

alanshaw:
Random idea that might lead to a reduction in time, we might want to tweak the batch size - it's [set to 16](https://github.com/storacha/blob-fetcher/blob/83137ba689068e37b86c441aae96658edd98762a/src/fetcher/batching.js#L16) but I think workers can only make 6 concurrent requests at a time (the public bucket is [fronted by a worker](https://github.com/storacha/public-bucket) for multipart byte range support but also because I figured in the future we might have some sort of UCAN auth to do...).

src/index.js
hannahhoward: note that if honeycomb api key is absent it sends nothing :) which for prod, for the timebeing it is :)

hannahhoward: we need to have a discussion about paying for honeycomb in general before we do anything other than turn on for dev testing.

hannahhoward:
@alanshaw I believe the spans only cover the time it takes to call fetch, and not the amount it takes to consume the body of the request. I suspect that's the time required to actually download the data.

PR #126, opened 2024-11-04
Egress is authorized with UCAN (stubbed)

Peeja:
Implements the UCAN validation for egress, except: can't yet get Spaces from the Indexing Service or get/store delegations, but both are stubbed so we can build from here.

## To test/use:

1. Start with a Freeway URL, like `https://bafybeib7l5an3dsnr65gvei4n3x64ihlqkhg4iytcrlyxkbse6m5e6zufm.ipfs.w3s.link`.
    * To test on localhost, convert from a subdomain gateway URL to a path gateway URL—eg, `http://localhost:8787/ipfs/bafybeib7l5an3dsnr65gvei4n3x64ihlqkhg4iytcrlyxkbse6m5e6zufm`
2. Grab the Space's DID. (This doesn't actually have to be the correct Space, as long as it's consistent, since we're stubbing the Space lookup *and* the delegation store.)
3. Make sure you're logged into the `w3` CLI.
4. Run `node scripts/delegate-serve.js <space-did> | pbcopy` to copy a `base64url` delegation string.
    * Or, use `node scripts/delegate-serve.js <space-did> <token> | pbcopy` to require a token.
5. Add `?stub_space=<space-did>&stub_delegation=<delegation>` to the URL.
    * Add `authToken=<token>` if a token was required.
6. The content should load!
    * If the token does not match, it shouldn't load.
    * If the Space does not match, it shouldn't load.
    * If the delegation is missing, it shouldn't load.
    * If additional delegations are provided, it should still load.

## The Stubs

* **`stub_space`:** The Space DID to "find" the content in. The lookup will really locate the content, but will add the given Space to whatever results it gets back.
* **`stub_delegation`:** A delegation to add to the "delegation store" for the request, `.archive()`ed to a CAR and `base64url`ed. May be given multiple times to add multiple delegations. Every call to `delegationsStore.find()` will find all given delegations.

## Remaining
- [x] Update to new ability names and wildcard delegation


src/middleware/withAuthorizedSpace.js
```diff
@@ -0,0 +1,192 @@
...
+ *
+ * @param {Ucanto.DID} space
+ * @param {AuthTokenContext & DelegationsStorageContext} ctx
+ * @returns {Promise<Ucanto.Result<{}, Ucanto.Failure>>}
+ */
+const authorize = async (space, ctx) => {
+  // Look up delegations that might authorize us to serve the content.
+  const relevantDelegationsResult = await ctx.delegationsStorage.find({
+    audience: ctx.gatewayIdentity.did(),
+    can: 'space/content/serve',
```
fforbeck, line 137: **suggestion**: Import definition from capability:
```suggestion
    can: [serve.can],
```

test/unit/middleware/withAuthorizedSpace.spec.js
```diff
@@ -0,0 +1,544 @@
...
+      )
+    )
+
+    expect(handler.notCalled).to.be.true
+    expectToBeInstanceOf(error, HttpError)
+    expect(error.status).to.equal(404)
+    expect(error.message).to.equal('Not Found')
+  })
+
+  it('should serve a found CID when stored in multiple Spaces', async () => {
```
fforbeck, line 240: ❤️ Excellent

src/middleware/withAuthorizedSpace.js
Peeja: Great call!

src/capabilities/serve.js
```diff
@@ -0,0 +1,51 @@
+import { capability, Schema, DID, nullable, string } from '@ucanto/validator'
+
+/**
+ * "Manage the serving of content owned by the subject Space."
+ *
+ * A Principal who may `space/content/serve/*` is permitted to perform all
+ * operations related to serving content owned by the Space, including actually
+ * serving it and recording egress charges.
+ */
+export const star = capability({
```
fforbeck, line 10: **suggestion**: I've released the new version of the `w3up` capabilities package in case you want to use the new `space/content/serve/*` from that lib.

Peeja: Ooh, excelent, thanks!

PR #127, opened 2024-11-06
fix(test): enable nodejs compat for miniflare

fforbeck:
#### **Overview**
This PR addresses an issue with resolving Node.js built-in modules (`node:buffer`, `node:events`, `node:async_hooks`) in a Cloudflare Worker Test environment. The Worker encountered errors in resolving these built-in modules, which aren’t natively supported in Workers. 

This issue was introduced when the [Opentelemetry PR was merged](https://github.com/storacha/freeway/pull/125), but we didn't see that because the Github Action that runs the `test:miniflare` was not flagging the test failure. 

#### **Error Description**
During testing, the following error appeared:
```
Unable to resolve "dist/worker.mjs" dependency "node:buffer": no matching module rules.
If you're trying to import a Node.js built-in module, or an npm package that uses Node.js built-ins, you'll either need to:
- Bundle your Worker, configuring your bundler to polyfill Node.js built-ins
- Configure your bundler to load Workers-compatible builds by changing the main fields/conditions
- Enable the `nodejs_compat` compatibility flag and use the `NodeJsCompatModule` module type
- Find an alternative package that doesn't require Node.js built-ins
```

#### **Solution**

1. To resolve this, I've enabled the `nodejs_compat` compatibility flag in the `test/miniflare/freeway.spec.js` file - which allows the Cloudflare Miniflare Worker to utilize Node.js runtime APIs, and set the compatibility date to "2024-09-23", as suggested in their docs: https://developers.cloudflare.com/workers/runtime-apis/nodejs/#get-started

2. Updated the Github Action to halt the execution in case of test failures

3. Updated the wrangler package to clean up some warnings during the integration tests.

#### **Note**
As of September 23, 2024, enabling `nodejs_compat` applies the same behavior as `nodejs_compat_v2` for compatibility dates set to September 23, 2024 or later, meaning this configuration is fully aligned with the latest runtime support for Node.js compatibility in Workers.

### DIFF
--- a/.github/actions/test/action.yml
+++ b/.github/actions/test/action.yml
@@ -13,12 +13,27 @@ runs:
       shell: bash
     - run: npm run lint
       shell: bash
-    - run: npm run test:unit
+    - run: |
+        npm run test:unit
+        if [ $? -ne 0 ]; then
+          echo "Unit Tests failed"
+          exit 1
+        fi
       name: Unit Tests
       shell: bash
-    - run: npm run test:miniflare
-      name: Miniflare Tests
-      shell: bash
-    - run: npm run test:integration
+    - run: |
+        npm run test:integration
+        if [ $? -ne 0 ]; then
+          echo "Integration Tests failed"
+          exit 1
+        fi
       name: Integration Tests
+      shell: bash
+    - run: |
+        npm run test:miniflare
+        if [ $? -ne 0 ]; then
+          echo "Miniflare Tests failed"
+          exit 1
+        fi
+      name: Miniflare Tests
       shell: bash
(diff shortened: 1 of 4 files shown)


.github/actions/test/action.yml
```diff
@@ -16,9 +16,18 @@ runs:
...
-    - run: npm run test:miniflare
-      name: Miniflare Tests
-      shell: bash
     - run: npm run test:integration
       name: Integration Tests
       shell: bash
+    - run: |
+        # If the Miniflare tests fail we need to capture the output and check for the error,
+        # because the error is not bubbled up from miniflare and the tests will pass.
+        npm run test:miniflare | tee test-output.log
```
Peeja, line 25: **suggestion (non-blocking):** Can we get `npm run test:miniflare` itself to `exit 1`? Ideally it shouldn't cover up failures at the command line either. If that's too complex for right now, let's add a card for it in the backlog.

Peeja (approved):
the simpsons flare

PR #130, opened 2024-11-22
fix(egress-client): set nonce, expire and fix servedAt

fforbeck:
Added the `nonce` as suggested per @alanshaw, but also set it to not expire, so we can process the invocation any time. The `servedAt` field doesn't need to be converted to seconds.

Blocked by https://github.com/storacha/w3infra/pull/443 & https://github.com/storacha/w3up/pull/1588

### DIFF
--- a/src/middleware/withEgressClient.js
+++ b/src/middleware/withEgressClient.js
@@ -89,8 +89,10 @@ async function record (space, resource, bytes, servedAt, env, ctx) {
     nb: {
       resource,
       bytes,
-      servedAt: Math.floor(servedAt.getTime() / 1000)
+      servedAt: servedAt.getTime()
     },
+    expiration: Infinity, // Don't expire the invocation, so we can record egress any time
+    nonce: process.hrtime().toString(),
     proofs: ctx.delegationProofs
   })
   const res = await invocation.execute(connection)


travis (approved):

src/middleware/withEgressClient.js
```diff
@@ -89,8 +89,10 @@ async function record (space, resource, bytes, servedAt, env, ctx) {
     nb: {
       resource,
       bytes,
-      servedAt: Math.floor(servedAt.getTime() / 1000)
+      servedAt: servedAt.getTime()
     },
+    expiration: Infinity, // Don't expire the invocation, so we can record egress any time
```
alanshaw, line 94: I'm not sure this is necessary, and I don't think the upload service should be able to record egress for this resource forever...just when it receives this invocation.

src/middleware/withEgressClient.js
```diff
@@ -89,8 +89,10 @@ async function record (space, resource, bytes, servedAt, env, ctx) {
...
+    nonce: process.hrtime().toString(),
```
alanshaw, line 95: If the partition key is now milliseconds precision we should probably also use that here right? At the moment we'd end up with multiple invocations recorded but only a single row in the DB (in the same millisecond). Using `Date.now()` here would mean we'd see the same number of invocations as DB rows (i.e. one in the same millisecond).

src/middleware/withEgressClient.js
fforbeck: I wanted to include this here explicitly because I was thinking about the rollout for Phase 2 (https://github.com/storacha/project-tracking/issues/192). In this phase, I plan to integrate a queue in Cloudflare worker to receive the egress events, with consumers responsible for signing and forwarding these requests to the Upload-API. This setup would allow us to fine-tune the duration for which these events are recorded, likely based on the throughput of the consumers. Also, this code would probably be moved to the consumer.

src/middleware/withEgressClient.js
fforbeck: @alanshaw I'm not sure I fully understand. My assumption was that in the context of generating a nonce, `process.hrtime()` would provide a more precise and unique value, particularly in a high-frequency execution environment. Does it need to align with the `servedAt` field (using Date.now())? I guess they intended to function independently, no?

PR #133, opened 2024-12-03
feat: ucan invocation handler

fforbeck:
### Context
To enable the gateway to serve content from a specific space, the space owner must delegate the `space/content/serve/*` capability to the Gateway. This delegation ensures the Gateway has the authority to serve the content and log egress events accurately.

This PR introduces a new handler to process POST requests to the server's root path. The handler acts as a UCAN Invocation handler, processing `access/delegate` invocations and extracting relevant delegation proofs. If a delegation proof is valid, it is stored in Cloudflare KV, allowing other handlers to retrieve and verify the proof to determine whether content should be served and egress logged.

Note: It doesn't cover the token verification.

#### **New Functionality**
- Added `withUcanInvocationHandler.js` to process `access/delegate` invocations:
  - Validates delegation proofs.
  - Stores valid proofs in a Cloudflare KV namespace dedicated to `content serve` delegations.
- Feature Flag: `FF_DELEGATIONS_STORAGE_ENABLED` if enabled, the new `withDelegationsStorage.js` handler will be used to find delegations in KV, and the existing `withDelegationsStubs.js` will be disabled.

### Related Issues
- https://github.com/storacha/project-tracking/issues/158
- https://github.com/storacha/project-tracking/issues/160


src/middleware/withAuthorizedSpace.types.ts
```diff
@@ -1,44 +1,16 @@
...
-}
-
-export interface DelegationsStorage {
-  /**
-   * find all items that match the query
-   */
-  find: (
-    query: Query
-  ) => Promise<Ucanto.Result<Ucanto.Delegation[], Ucanto.Failure>>
+  space?: SpaceDID
```
alanshaw, line 15: Would be good to get a comment here about what it means to have `space` (optionally) in the context. Something like if `space` is set it has been determined as the egress source for the response data.

src/server/utils.js
```diff
@@ -0,0 +1,72 @@
...
+ * Checks if the space/content/serve/* delegation is for the gateway and it is not expired.
+ *
+ * @param {import('@ucanto/interface').Signer} gatewayIdentity - The signer of the gateway identity
+ * @param {import('@ucanto/interface').InferInvokedCapability<typeof AccessCapabilities.delegate>} capability - The capability to validate
+ * @param {import('@ucanto/interface').Proof[]} proofs - The proofs to validate
+ */
+export const extractContentServeDelegation = (gatewayIdentity, capability, proofs) => {
+  const nbDelegations = new Set(Object.values(capability.nb.delegations))
+  if (nbDelegations.size > 1) {
+    return { error: new DelegationFailure(`nb.delegations has more than one delegation`) }
```
alanshaw, line 19: > `DelegationFailure`

How about `InvalidDelegation` for all these as what we're validating here is that the delegation we received is valid per our criteria?

src/server/index.js
```diff
@@ -0,0 +1,99 @@
...
+ * @template {import('@ucanto/interface').Caveats} C
+ * @template {{}} O
+ * @template {import('@ucanto/interface').Failure} X
+ * @template {import('@ucanto/interface').Result<O, X>} Result
+ * @param {import('@ucanto/interface').CapabilityParser<import('@ucanto/interface').Match<import('@ucanto/interface').ParsedCapability<A, R, C>>>} capability
+ * @param {(input:import('@ucanto/server').ProviderInput<import('@ucanto/interface').ParsedCapability<A, R, C>>) => import('@ucanto/interface').Await<Result>} handler
+ * @param {import('../middleware/withUcanInvocationHandler.types.js').Context} ctx
+ * @returns {import('@ucanto/interface').ServiceMethod<import('@ucanto/interface').Capability<A, R, C>, O & Result['ok'], X & Result['error']>}
+ */
+export const provide = (capability, handler, ctx) =>
```
alanshaw, line 43: You don't need this, just use `provide` from Ucanto server.

src/middleware/withDelegationsStorage.js
```diff
@@ -0,0 +1,80 @@
...
+        const value = await delegation.archive()
+        if (value.error) return value
+
+        await env.CONTENT_SERVE_DELEGATIONS_STORE.put(space, value.ok.buffer, options)
+        return { ok: {} }
+      } catch (error) {
+        const message = `error while storing delegation for space ${space}`
+        console.error(message, error)
+        return {
+          error: new DelegationFailure(message)
```
alanshaw, line 75: In w3infra we use `StoreOperationFailed` for this type of error.

src/middleware/withDelegationsStorage.types.ts
```diff
@@ -0,0 +1,48 @@
...
+  GatewayIdentityContext {
+  delegationsStorage: DelegationsStorage
+}
+
+export interface DelegationsStorage {
+  /**
+   * Finds the delegation proofs for the given space
+   * 
+   * @param {import('@web3-storage/capabilities/types').SpaceDID} space 
+   * @returns {Promise<Ucanto.Result<Ucanto.Delegation<Ucanto.Capabilities>, Ucanto.Failure>>}
```
alanshaw, line 31: 🙏 can we be more specific with the failure type? i.e. `DelegationNotFound | StoreOperationFailed`

src/server/utils.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+  const proofDelegations = proofs.flatMap((proof) => 'capabilities' in proof ? [proof] : [])
+  if (nbDelegations.size > proofDelegations.length) {
+    return { error: new DelegationFailure(`nb.delegations has more delegations than proofs`) }
```
alanshaw, line 23: Redundant as the check below will fail to find some of the delegations in proofs.

src/server/utils.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  const delegationLink = Array.from(nbDelegations)[0]
```
alanshaw, line 26: You might want to check there's only 1 and error otherwise.

src/server/utils.js
```diff
@@ -0,0 +1,72 @@
...
+  const delegationProof = proofDelegations.find((p) =>
+    delegationLink.equals(p.cid)
+  )
+  if (!delegationProof) {
+    return { error: new DelegationFailure(`no delegation found`) }
```
alanshaw, line 31:
```suggestion
    return { error: new DelegationFailure(`delegation not found in proofs: ${p.cid}`) }
```

src/server/utils.js
```diff
@@ -0,0 +1,72 @@
...
+
+  if (delegationProof.expiration && delegationProof.expiration !== Infinity && delegationProof.expiration < Math.floor(Date.now() / 1000)) {
+    return {
+      error: new DelegationFailure(
+        `delegation expired at ${delegationProof.expiration}`
+      )
+    }
+  }
+
+  if (delegationProof.capabilities.every((c) => c.can !== SpaceCapabilities.contentServe.can)) {
```
alanshaw, line 50: You just need one right? `some` not `every`?

test/unit/middleware/withAuthorizedSpace.spec.js
```diff
@@ -118,18 +120,18 @@ const gatewayIdentity = gatewaySigner.withDID('did:web:test.w3s.link')
...
-      )
-    })
-  })
+  find: async (space) => {
+    const delegation = delegations.find((d) =>
+      d.capabilities.some((cap) => cap.with === space)
+    )
+
+    if (!delegation) {
+      return { error: new Error('No matching delegation found') }
```
alanshaw, line 129:
```suggestion
      return { error: new DelegationNotFound() }
```

src/middleware/withDelegationsStorage.js
```diff
@@ -0,0 +1,80 @@
...
+        if (delegation.expiration && delegation.expiration > 0 && delegation.expiration !== Infinity) {
+          // expire the key-value pair when the delegation expires (seconds since epoch)
+          options = { expiration: delegation.expiration }
+        }
+
+        const value = await delegation.archive()
+        if (value.error) return value
+
+        await env.CONTENT_SERVE_DELEGATIONS_STORE.put(space, value.ok.buffer, options)
+        return { ok: {} }
```
alanshaw, line 70: Note: you can `import { ok, error } from '@ucanto/server'` to give you a better inferred return type (a Result).

```js
return ok({})
```

src/middleware/withDelegationsStorage.js
```diff
@@ -0,0 +1,80 @@
...
+     * @param {import('@web3-storage/capabilities/types').SpaceDID} space 
+     * @returns {Promise<Ucanto.Result<Ucanto.Delegation<Ucanto.Capabilities>, Ucanto.Failure>>}
+     */
+    find: async (space) => {
+      if (!space) return { error: { name: 'MissingSpace', message: 'No space provided' } }
+      const delegation = await env.CONTENT_SERVE_DELEGATIONS_STORE.get(space, 'arrayBuffer')
+      if (!delegation) return { error: { name: 'DelegationNotFound', message: `No delegation found for space ${space}` } }
+      const res = await Delegation.extract(new Uint8Array(delegation))
+      if (res.error) return res
+      return { ok: res.ok }
```
alanshaw, line 47:
```suggestion
      return Delegation.extract(new Uint8Array(delegation))
```

src/middleware/withDelegationsStorage.types.ts
```diff
@@ -0,0 +1,48 @@
...
+  find: (
+    space: SpaceDID
+  ) => Promise<Ucanto.Result<Ucanto.Delegation<Ucanto.Capabilities>, Ucanto.Failure>>
+
+  /**
+   * Stores the delegation proofs for the given space
+   * 
+   * @param {import('@web3-storage/capabilities/types').SpaceDID} space 
+   * @param {Ucanto.Delegation<Ucanto.Capabilities>} delegation
+   * @returns {Promise<Ucanto.Result<Ucanto.Unit, Ucanto.Failure>>}
```
alanshaw, line 42: 🙏 can we be more specific with the failure type? i.e. `StoreOperationFailed`

docs/ucan-handler.md
```diff
@@ -0,0 +1,44 @@
...
+1. **User Interaction**: The user interacts with the `w3up-client` to create a space and authorize the gateway to serve content.
+
+2. **UCAN Invocation**: The `w3up-client` invokes the `access/delegate` UCAN handler, providing the delegation details (`{ space, proofs }`). The request is sent to the Cloudflare `Access/Delegate Endpoint`.
+
+3. **Validation Steps**:
+   - The endpoint checks whether the space referenced in the delegation has been provisioned.
+   - It validates that the delegation matches the expected capability (`space/content/serve/*`).
+   - It ensures the proof chain is valid.
+
+4. **Relevance Check**: Only delegations associated with a provisioned space are accepted to prevent unnecessary resource usage and mitigate the risk of DoS attacks.
```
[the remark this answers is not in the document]
fforbeck: Yeah, I've added a comment saying we will tackle that in another moment.

src/middleware/withAuthorizedSpace.js
```diff
@@ -46,7 +49,7 @@ export function withAuthorizedSpace (handler) {
       ctx.authToken === null
 
     if (shouldServeLegacy) {
-      return handler(request, env, { ...ctx, space: null })
+      return handler(request, env, { ...ctx, space: undefined })
```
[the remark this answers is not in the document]
fforbeck: Not really. We were explicitly setting it to null | undefined because it’s a legacy request (one without associated delegations). I believe it’s safe to remove that now.

src/server/utils.js
fforbeck: That's true. Thanks

fforbeck:
> Note: as a follow up you'll need to update `reads` and `w3link` to proxy `POST /`.

Sounds good! The PR is ready for another review, @alanshaw .

src/middleware/withDelegationsStorage.js
```diff
@@ -0,0 +1,95 @@
...
+  get name() {
+    return /** @type {const} */ ('DelegationNotFound')
+  }
+}
+
+export class StoreOperationFailed extends Failure {
+  get name() {
+    return /** @type {const} */ ('StoreOperationFailed')
+  }
+}
```
alanshaw, line 26: You need to implement a `describe` method in these, unfortunately the way `Failure` is implemented causes a recursive call loop:

```js
import { Failure } from '@ucanto/server'

class DelegationNotFound extends Failure {
  get name() {
    return /** @type {const} */ ('DelegationNotFound')
  }
}

const err = new DelegationNotFound()
console.log(err.message)
```

```
RangeError: Maximum call stack size exceeded
    at DelegationNotFound.toString (<anonymous>)
    at DelegationNotFound.describe (file:///Users/alan/Code/storacha/w3up/node_modules/.pnpm/@ucanto+core@10.0.1/node_modules/@ucanto/core/src/result.js:65:17)
    at get message (file:///Users/alan/Code/storacha/w3up/node_modules/.pnpm/@ucanto+core@10.0.1/node_modules/@ucanto/core/src/result.js:68:17)
    at DelegationNotFound.toString (<anonymous>)
    at DelegationNotFound.describe (file:///Users/alan/Code/storacha/w3up/node_modules/.pnpm/@ucanto+core@10.0.1/node_modules/@ucanto/core/src/result.js:65:17)
    at get message (file:///Users/alan/Code/storacha/w3up/node_modules/.pnpm/@ucanto+core@10.0.1/node_modules/@ucanto/core/src/result.js:68:17)
    at DelegationNotFound.toString (<anonymous>)
    at DelegationNotFound.describe (file:///Users/alan/Code/storacha/w3up/node_modules/.pnpm/@ucanto+core@10.0.1/node_modules/@ucanto/core/src/result.js:65:17)
    at get message (file:///Users/alan/Code/storacha/w3up/node_modules/.pnpm/@ucanto+core@10.0.1/node_modules/@ucanto/core/src/result.js:68:17)
    at DelegationNotFound.toString (<anonymous>)
```

Something like this:


```js
class DelegationNotFound extends Failure {
  static name = /** @type {const} */ ('DelegationNotFound')
  #reason

  /** @param {string} [reason] */
  constructor (reason) {
    super()
    this.#reason = reason
  }

  get name() {
    return DelegationNotFound.name
  }

  describe() {
    return this.#reason ?? 'Delegation not found'
  }
}
```

Note: optional, but the static `name` property I added here allows users to switch on the error name without using magic string:

```js
if (res.error?.name === DelegationNotFound.name) {
  // do something with this specific error
}
```

src/middleware/withDelegationsStorage.js
```diff
@@ -0,0 +1,95 @@
...
+      let options = {}
+      if (delegation.expiration && delegation.expiration > 0 && delegation.expiration !== Infinity) {
+        // expire the key-value pair when the delegation expires (seconds since epoch)
+        options = { expiration: delegation.expiration }
+      }
+      
+      const value = await delegation.archive()
+      if (value.error) {
+        console.error(`error while archiving delegation`, value.error)
+        return error(value.error)
```
alanshaw, line 82:
```suggestion
        return value
```

src/server/index.js
```diff
@@ -0,0 +1,24 @@
...
+  return Server.create({
+    id: ctx.gatewaySigner,
+    codec: CAR.inbound,
+    service,
+    catch: err => console.error(err),
+    // TODO: wire into revocations
+    validateAuthorization: () => ({ ok: {} }),
+    // @ts-expect-error - The type is not defined in the ucan package, but it supports the method.
+    resolveDIDKey: (did) => resolveDIDKey(did, ctx),
+    authorities: [ctx.gatewayIdentity]
```
alanshaw, line 22: I think you can remove this right? It is not used by ucanto.

alanshaw (changes requested):

alanshaw (approved):

Peeja (dismissed):

src/middleware/withDelegationsStorage.js
```diff
@@ -0,0 +1,136 @@
...
+ * @param {DelegationsStorageEnvironment} env
+ * @returns {import('./withDelegationsStorage.types.js').DelegationsStorage}
+ */
+function createStorage(env) {
+  return {
+    /**
+     * Finds the delegation proofs for the given space
+     * 
+     * @param {import('@web3-storage/capabilities/types').SpaceDID} space 
+     * @returns {Promise<Ucanto.Result<Ucanto.Delegation<Ucanto.Capabilities>, DelegationNotFound | Ucanto.Failure>>}
```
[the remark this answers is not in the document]
fforbeck: As discussed during our call last week, I've updated it to support multiple delegations. Thanks for flagging that.

test/unit/middleware/withDelegationsStorage.spec.js
```diff
@@ -0,0 +1,97 @@
...
+    const mockHandler = sinon.stub().callsFake((request, env, ctx) => ctx)
+    const request = new Request('http://example.com/')
+    const env = {
+      FF_DELEGATIONS_STORAGE_ENABLED: 'false',
+      CONTENT_SERVE_DELEGATIONS_STORE: kvStoreMock
+    }
+
+    await withDelegationsStorage(mockHandler)(request, env, {
+      ...ctx,
+      // @ts-expect-error - we are testing the case where delegationsStorage is not set
```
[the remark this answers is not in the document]
fforbeck: I agree that using `@ts-expect-error` is not ideal, but this is strictly for testing purposes. That said, what is the use case for having both the stub and real implementation enabled simultaneously? If the real implementation is enabled, the goal should be to retrieve delegations from the KV store, not the stub. This is why I believe it's crucial to add the check in `withDelegationStubs` to prevent the risk of having both enabled simultaneously, which could lead to the `delegationsStorage` being overridden and causing unexpected behavior.

src/server/utils.js
```diff
@@ -0,0 +1,51 @@
...
+ * Checks if the space/content/serve/* delegation is for the gateway and it is not expired.
+ *
+ * @param {import('@ucanto/interface').Signer} gatewayIdentity - The signer of the gateway identity
+ * @param {import('@ucanto/interface').InferInvokedCapability<typeof AccessCapabilities.delegate>} capability - The capability to validate
+ * @param {import('@ucanto/interface').Proof[]} proofs - The proofs to validate
+ */
+export const extractContentServeDelegation = (gatewayIdentity, capability, proofs) => {
+  const nbDelegations = new Set(Object.values(capability.nb.delegations))
+  if (nbDelegations.size !== 1) {
+    return { error: new InvalidDelegation(`nb.delegations has more than one delegation`) }
```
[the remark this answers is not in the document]
fforbeck: It was originally designed to handle only a single delegation, specifically `space/content/serve/*`. However, as we discussed, this check is no longer necessary.

src/server/utils.js
```diff
@@ -0,0 +1,51 @@
...
+
+  if (delegationProof.expiration && delegationProof.expiration !== Infinity && delegationProof.expiration < Math.floor(Date.now() / 1000)) {
+    return {
+      error: new InvalidDelegation(
+        `delegation expired at ${delegationProof.expiration}`
+      )
+    }
+  }
+
+  if (!delegationProof.capabilities.some((c) => c.can === SpaceCapabilities.contentServe.can)) {
```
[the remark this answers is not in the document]
fforbeck: I don't think this is necessary at the moment, primarily because our client currently only grants the `space/content/serve/*` delegation. In the future, we might consider adding this if we decide to update the client to delegate `transport/http` and `egress/record` independently. For now, we can simplify by deriving these two capabilities from the top-level capability.

src/server/utils.js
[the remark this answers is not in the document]
fforbeck: After giving it more thought, I’ve decided to leave it as is for now since I won’t be implementing the verification for `transport/http` and `egress/record` at this time. Since these can be derived from the top-level capability, we can adjust the code later if we decide to implement individual delegations for specific sub-capabilities.

src/server/index.js
```diff
@@ -0,0 +1,42 @@
...
+export function createServer (ctx, service) {
+  return Server.create({
+    id: ctx.gatewaySigner,
+    codec: CAR.inbound,
+    service,
+    catch: err => console.error(err),
+    // TODO: wire into revocations
+    validateAuthorization: () => ({ ok: {} }),
+    // @ts-expect-error - The type is not defined in the ucan package, but it supports the method.
+    resolveDIDKey: (did) => resolveDIDKey(did, ctx)
```
Peeja, line 22: **question:** When do we need to resolve the Gateway's DID? The Gateway is the audience of these delegations, not the issuer. It shouldn't have signed anything that needs verifying. Right? Or am I missing something?

fforbeck: Good catch. That's right. We only need it to resolve the DID keys when we want to validate the issuer, and the issuer is using a `did:web`. I just removed that.

fforbeck:
@Peeja, how about we merge this without the `token` validation for now? Implementing the token will require additional work beyond the scope of Phase 1 of the Egress Billing project. For more details, see [issue #213](https://github.com/storacha/project-tracking/issues/213).
I believe it would be something to tackle in another iteration.

Peeja (commented):
Apologies for being slow to get back on this, it looks great!

PR #136, opened 2024-12-17
feat(indexer): probabilistic feature flag

fforbeck:
### Context

Implemented a new logic to determine whether the "indexing service" should be enabled for a given request.

1. **Query Parameter Check**: The code checks if the request URL contains a query parameter `ff` with the value `indexing-service`. If this parameter is present, it indicates that the indexing service should be enabled for this request.

2. **Ramp-Up Probability**: If the query parameter is absent, the code uses a probabilistic approach to decide whether to enable the indexing service. This is done using a "ramp-up probability," which represents the likelihood of enabling the service. The probability is defined by the environment variable `FF_RAMP_UP_PROBABILITY`. A default value of 0 (or 0%) is used if this variable is not set.

3. **Random Selection**: The code uses `Math.random()` to generate a random number between 0 and 1. If this number is less than or equal to the ramp-up percentage, the indexing service is enabled.

### Related Issues
- https://github.com/storacha/project-tracking/issues/223


src/middleware/withLocator.js
```diff
@@ -46,3 +44,22 @@ export function withLocator (handler) {
...
+ *
+ * @param {Request} request
+ * @param {LocatorEnvironment} env
+ * @returns {boolean}
+ */
+function isIndexingServiceEnabled (request, env) {
+  const withIndexingServicesArg = new URL(request.url).searchParams
+    .getAll('ff')
+    .includes('indexing-service')
+  const percentage = env.FF_RAMP_UP_PERCENTAGE ? Number(env.FF_RAMP_UP_PERCENTAGE) : 0.0
```
Peeja, line 62: **suggestion:** Let's either make `FF_RAMP_UP_PERCENTAGE` be 0–100, or change the name to not use "percentage"—maybe "probability"? I think the clearest is probably to use 0–100. We usually talk about it as a percentage, so we'll be setting it "to 50%", rather than "to 0.5".

Peeja (approved):
But What Are the Odds of That Happening?

LGTM, with one tweak.

src/middleware/withLocator.js
fforbeck: sounds good @Peeja . I will update that. Thanks

PR #137, opened 2024-12-20
fix: handle options req for cors preflight

fforbeck:
When we use the Console App to create a space, it will send an OPTIONS, and then a POST request to the UCAN Invocation Server, which is behind the CORS handler, and that CORS Handler doesn't support OPTIONS requests for CORS preflight. I've added a new handler to process that OPTIONS request.


travis (approved):

hannahhoward (approved):
LGTM

PR #140, opened 2025-01-04
fix(freeway): enable tracing properly and resolve fetch issue

hannahhoward:
# Goals

in #131 we believed https://github.com/storacha/project-tracking/issues/176 was solved. however, it turns out updates to freeway actually just prevented tracing from getting enabled properly, and the underlying issue remained.

# Implementation

- bind globalthis.fetch to itself (this is the proper fix to https://github.com/storacha/project-tracking/issues/176)
- properly enable tracing

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -133,8 +133,11 @@ let handlerPromise = null
  */
 async function initializeHandler (env) {
   const baseHandler = middleware(handleUnixfs)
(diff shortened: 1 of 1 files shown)


fforbeck (approved):
Good catch!! Thanks for fixing that. LGTM

PR #143, opened 2025-01-05
feat(tracing): enable setting a sampling ratio for tracing

hannahhoward:
# Goals

Set us up to enable traces in our prod w3s.link environment

# Implementation

Essentially, to enable tracing in prod, we need to cut down on number of requests we sample traces for. This PR introduces the simplest form of sampling: head sampling, where we simply only enable traces on a small set of samples

The next step would be to do some kind of tail sampling -- sent traces on error response codes for example.

For now I set the sampling ratio for prod to 1% -- could use some guidance from @travis who might have the best sense of how many requests we're talking about -- our current Honeycomb account supports 100 million per month.


src/bindings.d.ts
```diff
@@ -15,4 +15,5 @@ export interface Environment
   CONTENT_CLAIMS_SERVICE_URL?: string
   HONEYCOMB_API_KEY: string
   FF_TELEMETRY_ENABLED: string
+  FF_TELEMETRY_RATIO: string
```
alanshaw, line 18: Not a feature flag, but optional config.

```suggestion
  TELEMETRY_RATIO: string
```

src/index.js
```diff
@@ -109,7 +110,10 @@ function config (env, _trigger) {
         url: 'https://api.honeycomb.io/v1/traces',
         headers: { 'x-honeycomb-team': env.HONEYCOMB_API_KEY }
       },
-      service: { name: 'freeway' }
+      service: { name: 'freeway' },
+      sampling: {
+        headSampler: new TraceIdRatioBasedSampler(parseFloat(env.FF_TELEMETRY_RATIO))
+      }
```
alanshaw, line 116: Only enable sampling if ratio is set?

```suggestion
      ...(env.FF_TELEMETRY_RATIO ? {
        sampling: {
          headSampler: new TraceIdRatioBasedSampler(parseFloat(env.FF_TELEMETRY_RATIO))
        }
      } : {})
```

alanshaw (changes requested):

travis:
Free
Screenshot 2025-01-06 at 11 22 47 AM

alanshaw (approved):

hannahhoward:
@travis lowered the ratio to 0.0005 cause I now see that the billing is per span and I think we're looking at 20 spans or so per trace.

PR #147, opened 2025-01-23
feat(dag-protobuf): cache dag pb directory structure and block indexes

fforbeck:
### Context
The requests to fetch a DAG Protobuf directory structure using a CID execute the following steps:
1. Get all content claims lookups to identify where we are pulling data from
2. Fetch cid `bafy...cid` - which represents the folder containing the target file, so that we can determine the verifiable cid for the file (let's call that `bafy...file`)
3. Fetch cid `bafy...file` to get the root block of the file, which in UnixFS contains NO raw data, but rather is a list of sub-blocks that contain the file (let's call those `bafy...bytes1` and `bafy...bytes2`)
4. Fetch the first raw data blocks to send the first byte

This PR enables the caching strategy for steps 2 to 4 where instead of fetching the directory structure from the locator and navigating the DAG for every request, it caches the DAGs if they have a  Protobuf structure and content size <= 2MB.

### Changes
- Updated `withContentClaimsDagula` middleware to cache DAG PB content requests
- New KV Store
  - `DAGPB_CONTENT_CACHE`
- Caching rules
  - `FF_DAGPB_CONTENT_CACHE_TTL_SECONDS`: The number that represents when to expire the key-value pair in seconds from now. The minimum value is 60 seconds. Any value less than 60MB will not be used. We will use **30 days TTL** by default for Production environment.
  - `FF_DAGPB_CONTENT_CACHE_MAX_SIZE_MB`: The maximum size of the key-value pair in MB. The minimum value is 1 MB. Any value less than 1MB will not be used. We will use **2MB max file size** by default.
  - `FF_DAGPB_CONTENT_CACHE_ENABLED`: The flag that enables the DAGPB content cache. The cache is **disabled in prod** by default.

### Samples
**2MB file - no cache**
2mb-file-no-cache
- 5.2 seconds

**2MB file - cached**
2mb-file-cached
- 1.4 seconds

### KV Limits
- Reads: unlimited
- Writes (different keys): unlimited
- Writes (same key): 1w / sec (rate limiting)
- Storage/account & Storage/namespace: unlimited
- key size: <= 512 bytes
- value size: <= 25MiB
- Minimum cache ttl: 60 seconds
- Higher limit? -> https://forms.gle/ukpeZVLWLnKeixDu7

resolves https://github.com/storacha/project-tracking/issues/301


hannahhoward (approved):
LGTM though I think we should set longer TTL -- we can just drop expiry and see how large the cache gets over time. the big expense is writes anyway.

src/middleware/withContentClaimsDagula.js
```diff
@@ -46,3 +52,70 @@ export function withContentClaimsDagula (handler) {
...
+  if (env.FF_DAGPB_CONTENT_CACHE_ENABLED === 'true' && cid.code === dagPb.code) {
+    const cachedBytes = await getCachedDagPbBytes(env, cid)
+    if (cachedBytes) {
+      return { cid, bytes: cachedBytes }
+    }
+
+    const res = await fetcher.fetch(cid.multihash)
+    if (res.ok) {
+      const bytes = await res.ok.bytes()
+      const cachedBytes = await cacheDagPbBytes(env, cid, bytes)
```
alanshaw, line 75: Do in `waitUntil`?

src/middleware/withContentClaimsDagula.js
```diff
@@ -46,3 +52,70 @@ export function withContentClaimsDagula (handler) {
...
+ *
+ * @param {Environment} env
+ * @param {import('multiformats').UnknownLink} cid
+ * @param {Uint8Array} bytes
+ * @returns {Promise<Uint8Array>}
+ */
+async function cacheDagPbBytes (env, cid, bytes) {
+  if (env.FF_DAGPB_CONTENT_CACHE_MAX_SIZE_MB && bytes.length <= env.FF_DAGPB_CONTENT_CACHE_MAX_SIZE_MB * 1024 * 1024) {
+    try {
+      await env.DAGPB_CONTENT_CACHE.put(cid.multihash.digest.toString(), bytes.buffer, {
```
alanshaw, line 94: Hmm, I don't know that the underlying `buffer` will necessarily contain the bytes you want, i.e. the block bytes may have been sliced from a bigger chunk.

alanshaw, line 94: > cid.multihash.digest.toString()

I don't think that'll give you something good...

```js
> const Link = await import('multiformats/link')
undefined
>
> Link.parse('bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy').multihash.digest.toString()
'199,208,20,137,8,8,88,197,0,6,88,54,198,88,248,71,166,202,103,196,134,70,25,33,43,228,248,32,14,75,186,206'
```

You can base58 encode it?

```js
> const { base58btc } = await import('multiformats/bases/base58')
undefined
>
> base58btc.encode(Link.parse('QmNpuky6TiLHbcNbjn2tmnisTcfFVhhyeTCRrJns1NkDfK').multihash.bytes)
'zQmNpuky6TiLHbcNbjn2tmnisTcfFVhhyeTCRrJns1NkDfK'
>
> base58btc.encode(Link.parse('bafybeiahhvdwxvtwlakzq3rgze54z7ydg7py2f74xcp6ts77mxj2g4brmy').multihash.bytes)
'zQmNpuky6TiLHbcNbjn2tmnisTcfFVhhyeTCRrJns1NkDfK'
```

Note: `multihash.bytes` not `multihash.digest`.

alanshaw, line 94: Old DAGs will not have "raw" blocks as leaves, they will be `dag-pb` and the UnixFS data (decoded from the `Data` property of the node) will be of type "file". We should not cache these.

Screenshot 2025-01-24 at 11 28 13

I think you don't have to decode the unixfs data, you just need to `dagPb.decode(bytes)` and check that `Links` has length 0.

alanshaw (changes requested):

alanshaw (changes requested):
I'd set a TTL of a month or something.

Needs tests.

src/middleware/withContentClaimsDagula.js
fforbeck: @alanshaw what would be the best option then?

fforbeck: Got it. I will change that. thx

fforbeck: Nice. Thank you!

test/miniflare/freeway.spec.js
```diff
@@ -49,17 +49,26 @@ describe('freeway', () => {
     const { port } = server.address()
     url = new URL(`http://127.0.0.1:${port}`)
     miniflare = new Miniflare({
+      host: '127.0.0.1',
+      port: 8787,
+      inspectorPort: 9898,
+      log: new Log(LogLevel.INFO),
+      cache: false, // Disable Worker Global Cache to test cache middlewares
```
fforbeck, line 56: If we don't disable this global cache in test the requests won't reach the Content Claim Dagula middleware where we have the KV cache.

src/middleware/withContentClaimsDagula.js
alanshaw: I *think* slice (take a copy) and then access the `.buffer`

test/miniflare/freeway.spec.js
```diff
@@ -481,4 +495,88 @@ describe('freeway', () => {
...
+    console.log('SECOND REQUEST')
+    const res2 = await miniflare.dispatchFetch(`http://localhost:8787/ipfs/${root}/${input[2].name}`, {
+      headers: {
+        'Cache-Control': 'no-cache'
+      }
+    })
+    if (!res2.ok) assert.fail(`unexpected response: ${await res2.text()}`)
+    const end2 = performance.now()
+    assertBlobEqual(input[2], await res2.blob())
+    assert(end2 - start2 < end - start, 'Second request should take less time than the first request')
```
hannahhoward, line 549: timing tests are a bit unreliable. let's monitor and address if this becomes a flaky test.

wrangler.toml
```diff
@@ -53,9 +54,12 @@ MAX_SHARDS = "825"
 FF_RATE_LIMITER_ENABLED = "false"
 FF_EGRESS_TRACKER_ENABLED = "true"
 FF_TELEMETRY_ENABLED = "true"
-TELEMETRY_RATIO = 0.0005
 FF_DELEGATIONS_STORAGE_ENABLED = "true"
 FF_RAMP_UP_PROBABILITY = "0"
+FF_DAGPB_CONTENT_CACHE_TTL_SECONDS = 86400
```
hannahhoward, line 59: @alanshaw suggested making it 30 days?

hannahhoward (approved):
LGTM on condition we set the prod cache to 30 days.

test/miniflare/freeway.spec.js
fforbeck: sounds good.

wrangler.toml
fforbeck: That's right. Just fixed. Thanks for flagging.

src/middleware/withContentClaimsDagula.js
fforbeck: I ended up using the actual bytes, and that worked well.

PR #150, opened 2025-01-30
feat(deps): update to new blob-fetcher

hannahhoward:
# Goals

Use the new refactor freeway

# Implementation

- update dependencies for blob-fetcher & indexing-service-client
- update withLocator to use new initializations
- change on the factories for location claims in the tests -- this is weird cause normally I'd say tests breaking is a bad sign, but this certianly appears to be a genuine bug in the factor that was covered up by a potential bug in the old code (always using digest as the location claim's key even if content in the actual claim was different)


fforbeck (approved):

Peeja (approved):
LGTM!

PR #173, opened 2025-05-16
Add FF to turn *off* Indexing Service for debugging

Peeja:
Pushed to debug issue with retrieval, but may not be necessary in the end.


Peeja:
Force-pushed a formatting fix

PR #174, opened 2025-05-16
fix: handling legacy requests

fforbeck:
### Context

1. The content was uploaded before we implemented the current indexing and carpark system
2. Our current gateway tries to use the indexing service if enabled, if not, it will try to find in the carpark directly, but for some reason, it can't find it there as well
3. When neither system can find the content, the request fails, but if we disable both services, it will use content claims, and then it will find the correct location in the R2 carpark

### Solution
Add a `legacyReq=true` query parameter that allows bypassing both the indexing service and carpark systems. When this parameter is present:
- The request will skip the indexing service check
- The carpark functionality will be disabled
- The content will be resolved via the content claims service

This change has been implemented in two key places:
1. `withLocator.js`: Disables both indexing service and carpark when `legacyReq=true`
2. `withCarParkFetch.js`: Bypasses carpark fetch functionality when `legacyReq=true`

### DIFF
--- a/src/middleware/withCarParkFetch.js
+++ b/src/middleware/withCarParkFetch.js
@@ -54,10 +54,12 @@ const MAX_BATCH_SIZE = 20 * 1024 * 1024
  */
 export function withCarParkFetch (handler) {
   return async (request, env, ctx) => {
-    // if carpark public bucket is not set, just use default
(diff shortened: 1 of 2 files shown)


fforbeck:
I'll merge this hotfix and deploy. Happy to open other PRs to solve issues.

